PR #1, opened 2020-08-13
update storybook + dependencies & add missing dependencies to do semantic-release

larsdouweschuitema:
- Updating storybook and its dependencies to the new version. Tested locally and works fine.
- Add missing dependencies semantic-release

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1433,6 +1433,51 @@
                 "react-lifecycles-compat": "^3.0.4"
             }
         },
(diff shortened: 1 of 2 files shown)


jpsc:
Could you please separate the update of the storybook from the semantic release fix?
Semantic release still needs to be properly configured.

larsdouweschuitema:
> Could you please separate the update of the storybook from the semantic release fix?
> Semantic release still needs to be properly configured.

Done

PR #25, opened 2020-08-28
Feature/tailwindcss

jpsc:
Add tailwind config and button examples


src/stories/ui-button.stories.js
```diff
@@ -0,0 +1,41 @@
+import UiButtonPrimary from '@/components/ui-button-primary.vue';
```
larsdouweschuitema, line 1: @jpsc There are different ways of defining these files. See for example the open PR of the button I wrote. I have experimented with both ways. We should talk shortly about what patterns we are going to stick with.
Not sure if you did already, but might be worth checking out @storybook/vue.

Also not sure about having so many files for a single component. It's bloat I think.

src/styles.css
```diff
@@ -0,0 +1,21 @@
+@font-face {
+    font-family: 'Proxima Nova';
```
larsdouweschuitema, line 2: These properties can be on a single line just using the "font" property.

.storybook/main.js
```diff
@@ -1,8 +1,4 @@
 module.exports = {
     stories: ['../src/**/*.stories.js'],
-    addons: [
-        '@storybook/addon-knobs/register',
-        '@storybook/addon-notes/register',
-        '@storybook/addon-storysource',
-    ]
+    addons: ['@storybook/addon-essentials'],
```
larsdouweschuitema, line 3: Nice. I did not know about this addon containing all essentials. 👍

.storybook/preview.js
```diff
@@ -0,0 +1 @@
+import './../src/styles.css'
```
larsdouweschuitema, line 1: Use alias @

postcss.config.js
```diff
@@ -0,0 +1,5 @@
+module.exports = {
+    plugins: [
+      require('tailwindcss')
+    ]
+  }
```
larsdouweschuitema, line 5: Newline

.storybook/main.js
jpsc: It's new since Storybook v6

postcss.config.js
jpsc: Let's fix that automatically when we have configured prettier.

src/styles.css
jpsc: Indeed it can but i've always found that its hard to know the correct order of all the properties on the `font` property.
I don't think this is a big issue since it's just for storybook.

dine (approved):

package.json
```diff
@@ -24,11 +24,13 @@
     "devDependencies": {
         "@babel/core": "^7.11.4",
         "@babel/preset-env": "^7.11.0",
+        "@funda/tailwind-config": "^1.1.0",
```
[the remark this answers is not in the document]
jpsc: @dine you can add the button's custom box-shadow to the tailwind config on this repo and then if makes sense later on move it to the funda config.

src/stories/ui-button.stories.js
jpsc: > Also not sure about having so many files for a single component. It's bloat I think.
I agree. I was just testing. Let's fix this.

postcss.config.js
dine: just pushed prettier, you can get the changes and check 🙃

larsdouweschuitema (approved):

PR #55, opened 2020-09-29
Notification component

bartderijk:
I wanted to expose this one early, so there's still some work left one this one but I'd like to get feedback so I can start iterating right away.


src/components/ui-notification.vue
```diff
@@ -0,0 +1,156 @@
...
+                    <rect
+                        x="22.5"
+                        y="32"
+                        width="3"
+                        height="3"
+                        rx="1.5"
+                        ry="1.5"
+                        transform="translate(48 67) rotate(-180)"
+                    />
+                    <path
```
dine, line 47: I thought we would use funda-icons here, should we? 🤔

dine (commented):
I also noticed the commit messages are not following the https://semver.org/ convention, so we won't get a version bump after merging this
[update] just noticed it's not even working

larsdouweschuitema:
Add test

src/components/ui-notification.vue
bartderijk: As discussed, funda-icons is currently not available here. We'll add them separately from this PR, and replace the hardcoded icons here soon as that's finished.

bartderijk:
> Add test

@larsdouweschuitema I'm all for testing functionality, but this component has basically nothing functional. What part would you like to add a test for?

dine (approved):
We can update to use funda-icons later, but I believe that we need to fix the semantic-release actions before the next component

bartderijk:
> We can update to use funda-icons later, but I believe that we need to fix the semantic-release actions before the next component

That makes sense, let's take that approach 👍🏻

PR #112, opened 2020-10-29
Toast component

bartderijk:
I've added the toast component, it's styles are directly converted so it should visually be exactly the same as it was previously. I was thining of adding tests, but then I'm not sure what to test on this one, I'm curious what you think here.

toast

### DIFF
--- a/src/components/ui-toast.vue
+++ b/src/components/ui-toast.vue
@@ -123,6 +123,7 @@
                 <ui-button
                     tone="tertiary"
                     class="ui-top-0 ui-right-0 ui-absolute ui-px-2 ui-opacity-0 group-hover:ui-opacity-100"
+                    @click="close"
                 >
                     <span class="ui-w-6 ui-h-6" aria-hidden="true">
                         <svg
@@ -168,6 +169,11 @@ export default {
                 success: 'ui-border-green-1',
                 warning: 'ui-border-yellow-1',
             }[this.type];
+        },
+    },
+    methods: {
+        close() {
+            this.$emit('close');
         },
     },
 };
--- a/src/components/ui-toast.vue
+++ b/src/components/ui-toast.vue
@@ -2,137 +2,142 @@
     <div
         class="ui-fixed ui-top-0 ui-right-0 ui-overflow-x-hidden ui-z-175 ui-max-w-sm"
     >
-        <div
-            class="ui-group ui-shadow ui-relative ui-rounded-sm ui-bg-white ui-border-solid ui-border-l-4 ui-flex ui-p-4 ui-pr-10 ui-m-3"
-            :class="[borderColorClass, animationClass]"
-        >
-            <div class="ui-mr-3" :aria-label="type">
-                <svg
-                    v-if="type === 'information'"
-                    class="ui-align-middle ui-w-6 ui-h-6 ui-text-blue-2"
-                    xmlns="http://www.w3.org/2000/svg"
-                    viewBox="0 0 48 48"
-                >
-                    <title>Information</title>
-                    <rect
-                        x="22.5"
-                        y="18"
-                        width="3"
-                        height="17"
-                        rx="1.5"
-                        ry="1.5"
-                    />
-                    <path d="M24,16h0a1.5,1.5,0,0,0,0-3h0a1.5,1.5,0,0,0,0,3Z" />
-                    <path
-                        d="M24,4A20,20,0,1,0,44,24,20,20,0,0,0,24,4Zm0,37A17,17,0,1,1,41,24,17,17,0,0,1,24,41Z"
-                    />
-                </svg>
-
-                <svg
-                    v-if="type === 'error'"
-                    class="ui-align-middle ui-w-6 ui-h-6 ui-text-red-1"
-                    xmlns="http://www.w3.org/2000/svg"
-                    viewBox="0 0 48 48"
-                >
-                    <title>Error</title>
-                    <g id="Error">
+        <transition name="animation--fade-right">
+            <div
+                v-if="show"
+                class="ui-group ui-shadow ui-relative ui-rounded-sm ui-bg-white ui-border-solid ui-border-l-4 ui-flex ui-p-4 ui-pr-10 ui-m-3"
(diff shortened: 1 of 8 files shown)


src/components/ui-toast.vue
```diff
@@ -0,0 +1,211 @@
...
+    computed: {
+        borderColorClass() {
+            return {
+                information: 'ui-border-blue-2',
+                error: 'ui-border-red-1',
+                success: 'ui-border-green-1',
+                warning: 'ui-border-yellow-1',
+            }[this.type];
+        },
+        animationClass() {
```
dine, line 169: what do you think about using transitions and the vue animation class names here?
https://vuejs.org/v2/guide/transitions.html

src/components/ui-toast.vue
```diff
@@ -0,0 +1,211 @@
...
+    50% {
+        opacity: 1;
+    }
+
+    100% {
+        transform: translateX(0);
+    }
+}
+
+svg:not([fill]) {
```
dine, line 208: I wish this would be default, like all css from tailwind

src/components/ui-toast.vue
```diff
@@ -0,0 +1,211 @@
...
+    },
+};
+</script>
+
+<style scoped>
+.toast__close {
+    transition: opacity 0.25s cubic-bezier(0.23, 1, 0.32, 1);
+}
+
+.toast__content:hover > .toast__close {
```
dine, line 184: could we replace by the class `hover:ui-opacity-100`?

src/components/ui-toast.vue
bartderijk: I like the way you think :) that would be cool! Looking into that, it could be a really nice addition.

src/components/ui-toast.vue
bartderijk: I don't think so, as it is the child of the hovered element that needs to get the opacity changed on hover. But I'll double check.

bartderijk: @dine I _should_ be possible via https://tailwindcss.com/docs/pseudo-class-variants#group-hover but I'm somehow it doesn't work for me, even after enabling in tailwind.config.js. Still diggin....

bartderijk: @dine Fixed!

src/components/ui-toast.vue
bartderijk: @dine Fixed! This is such a cool addition!

src/components/ui-toast.vue
```diff
@@ -0,0 +1,205 @@
...
+                success: 'ui-border-green-1',
+                warning: 'ui-border-yellow-1',
+            }[this.type];
+        },
+    },
+};
+</script>
+
+<style scoped>
+.animation--fade-right-enter-active {
```
sukalas-a, line 177: Shouldn't we ditch the traditional styling and do as much as possible using Tailwind. 
When that is not possible we can add the custom styles in the extended config.
Thoughts?

src/components/ui-toast.vue
```diff
@@ -0,0 +1,205 @@
...
+    props: {
+        type: {
+            type: String,
+            default: 'information',
+            validator: (value) =>
+                ['information', 'error', 'success', 'warning'].includes(value),
+        },
+        show: {
+            type: Boolean,
+            default: true,
```
sukalas-a, line 160: Shouldn't the default be false?

sukalas-a:
> I've added the toast component, it's styles are directly converted so it should visually be exactly the same as it was previously. I was thining of adding tests, but then I'm not sure what to test on this one, I'm curious what you think here.
> 
> toast

The only thing that could be tested is that when show is true the component is visible. 
So pretty much securing that no-none removes the v-if from the component.

src/components/ui-toast.vue
bartderijk: Vue animation needs the class name to post-fix `enter-active`/`leave-active` etc. That's something we can't get around if we want to benefit from Vue's transition functions, and I'm keep to have them. We could extend tailwind's animation definitions I suppose, I'll look into that!

src/components/ui-toast.vue
bartderijk: Fixed!

src/components/ui-toast.vue
bartderijk: Implemented it using https://tailwindcss.com/docs/animation#customizing However it isn't working yet, looking into this now..

src/components/ui-toast.vue
```diff
@@ -0,0 +1,205 @@
...
+        class="ui-fixed ui-top-0 ui-right-0 ui-overflow-x-hidden ui-z-175 ui-max-w-sm"
+    >
+        <transition name="animation--fade-right">
+            <div
+                v-if="show"
+                class="ui-group ui-shadow ui-relative ui-rounded-sm ui-bg-white ui-border-solid ui-border-l-4 ui-flex ui-p-4 ui-pr-10 ui-m-3"
+                :class="[borderColorClass]"
+            >
+                <div class="ui-mr-3" :aria-label="type">
+                    <svg
```
daniarsuprianto, line 12: can we use funda icons here?

daniarsuprianto: i meant, for all icons in this file

src/components/ui-toast.vue
bartderijk: Fixed! Thanks for the feedback!

src/components/ui-toast.vue
bartderijk: For sure! My next addition to the lib will be the inclusion of icons. Soon as that's available I'll be sure to replace the svg's here with funda icons.

sukalas-a:
Looks good! 👍

sukalas-a (approved):
Awesome that we got rid of the scoped styling, looks clean!

src/components/ui-toast.vue
bartderijk: @dine Guess what @sukalas-a came up with: `ui-fill-current` does exactly that!

PR #127, opened 2020-11-04
Add ui-select component

renancarvalho:
Adds the select component.

### DIFF
--- a/src/components/ui-select.vue
+++ b/src/components/ui-select.vue
@@ -3,8 +3,9 @@
         class="ui-relative ui-container ui-rounded-sm ui-rounded-sm ui-border ui-border-light-1 ui-bg-white ui-cursor-pointer ui-shadow-sm"
     >
         <select
-            class="ui-container ui-cursor-pointer ui-outline-none ui-text-base ui-font-normal ui-leading-normal ui-p-2 ui-pl-4"
-            @change="onChange"
+            class="ui-appearance-none ui-container ui-cursor-pointer ui-outline-none ui-text-base ui-font-normal ui-leading-normal ui-p-2 ui-pl-4"
+            v-bind="$attrs"
+            v-on="$listeners"
         >
             <option
                 v-for="option in options"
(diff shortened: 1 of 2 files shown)


src/components/ui-select.vue
```diff
@@ -0,0 +1,59 @@
...
+        onChange() {
+            this.$emit('change');
+        },
+    },
+};
+</script>
+<style scoped>
+select {
+    -webkit-appearance: none;
+    -moz-appearance: none;
```
dine, line 55: I think we should avoid using css on the components if possible.
https://tailwindcss.com/docs/appearance#class-reference

src/components/ui-select.vue
```diff
@@ -0,0 +1,59 @@
+<template>
+    <div
+        class="ui-relative ui-container ui-rounded-sm ui-rounded-sm ui-border ui-border-light-1 ui-bg-white ui-cursor-pointer ui-shadow-sm"
+    >
+        <select
+            class="ui-container ui-cursor-pointer ui-outline-none ui-text-base ui-font-normal ui-leading-normal ui-p-2 ui-pl-4"
+            @change="onChange"
```
dine, line 7: What do you think about forwarding all the attributes and listeners here?
```
 v-bind="$attrs"
 v-on="$listeners"
```

src/components/ui-select.vue
```diff
@@ -0,0 +1,59 @@
...
+                    data-name="Arrow Down"
+                    d="M35.66 18.83a1.5 1.5 0 0 0-2.12 0L24 28.37l-9.54-9.54A1.5 1.5 0 0 0 12.34 21l10.57 10.53a1.52 1.52 0 0 0 2.13 0L35.66 21a1.5 1.5 0 0 0 0-2.17z"
+                ></path>
+            </svg>
+        </div>
+    </div>
+</template>
+
+<script>
+export default {
```
dine, line 34: If we forward the attributes like I mentioned, we should avoid them going to the root element
```
inheritAttrs: false,
```

dine (commented):
Nice, just some small remarks

dine (dismissed):

dine (approved):

PR #128, opened 2020-11-04
Input component

renancarvalho:
Add input component


src/components/ui-input.vue
```diff
@@ -0,0 +1,42 @@
+<template>
+    <div class="ui-container">
```
dine, line 2: I think we should replace this class for `ui-relative`
For when we need to add prefix or suffix to input

src/components/ui-input.vue
```diff
@@ -0,0 +1,42 @@
...
+        <input
+            :type="type"
+            :value="value"
+            :class="isValid !== true ? 'ui-border-red-1' : 'ui-border-light-1'"
+            class="ui-container ui-rounded-sm ui-rounded-sm ui-border ui-bg-white ui-shadow-sm ui-p-2 ui-pl-4 ui-outline-none ui-text-base ui-font-normal"
+            :placeholder="placeHolder"
+            v-bind="$attrs"
+            v-on="$listeners"
+        />
+        <p v-if="!isValid" class="ui-text-red-1 ui-mt-1">
```
dine, line 12: I think this should be another component, like `ui-error-message`

src/components/ui-input.vue
```diff
@@ -0,0 +1,42 @@
+<template>
+    <div class="ui-container">
+        <input
+            :type="type"
+            :value="value"
```
dine, line 5: If you are binding attributes I don't think we need this line

src/components/ui-input.vue
```diff
@@ -0,0 +1,42 @@
...
+            :class="isValid !== true ? 'ui-border-red-1' : 'ui-border-light-1'"
+            class="ui-container ui-rounded-sm ui-rounded-sm ui-border ui-bg-white ui-shadow-sm ui-p-2 ui-pl-4 ui-outline-none ui-text-base ui-font-normal"
+            :placeholder="placeHolder"
```
dine, line 8: Same comment as before

src/components/ui-input.vue
renancarvalho: Indeed! added to PR too!

src/components/ui-input.vue
renancarvalho: "Type" didnt work without.
"Value: is gone

src/components/ui-input.vue
renancarvalho: Placeholder must be there event with the v-bind...

src/components/ui-input.vue
renancarvalho: Think we need both  =)

dine (approved):

PR #131, opened 2020-11-06
Remove ui prefix

bartderijk:
Following the bi-weekly call to remove the pre-fix, I adjusted the code base accordingly.


dine (approved):

daniarsuprianto (commented):
Are we going to do this also for all projects with tailwind in the end?

dine:
@daniarsuprianto I think it's safe, funda-ui has a different prefix ('fd-') and the chances of class names collision seem to be low in our projects, as we didn't use to do many functional classes there. Do you have concerns?

daniarsuprianto:
@dine nothing, i was just wondering because last time we discussed about the ui- prefix, and now we changed without prefix.. so just curious why and what's the reason. i dont see there'll be any conflict or whatsoever with current projects because other than funda-ui utilities classes, we dont work 'utility-first'

daniarsuprianto (approved):

renancarvalho:
NICE! no more ui-!
Good!

renancarvalho (approved):
-

larsdouweschuitema (approved):

larsdouweschuitema:
Thanks for the work on this @bartderijk

PR #140, opened 2020-11-12
feat(icon): add icon component

dine:


src/components/ui-notification.vue
```diff
@@ -1,110 +1,40 @@
...
+                class="text-yellow-1"
+            ></ui-icon>
         </span>
 
         <slot name="notificationBody" />
     </div>
 </template>
 
 <script>
+import UiIcon from './ui-icon.vue';
```
larsdouweschuitema, line 34: Are we still using aliases? Although this path is rather simple, it makes it cleaner.

src/stories/ui-icon.stories.js
```diff
@@ -0,0 +1,182 @@
...
+
+export default {
+    title: 'Informational/Icon',
+    component: UiIcon,
+    argTypes: {
+        icon: {
+            control: {
+                type: 'select',
+                options: [
+                    'addHouse',
```
larsdouweschuitema, line 11: Can we keep the export itself clean and move this inside a const?

src/stories/ui-notification.stories.js
```diff
@@ -29,19 +30,11 @@ const Template = (args, { argTypes }) => ({
 
                 <button
                     type="button"
-                    class="leading-6 rounded-sm px-6 h-11 inline-flex text-center items-center justify-center align-middle cursor-pointer whitespace-no-wrap ml-3 bg-transparent border-none p-0 h-auto border-0 text-dark-2"
+                    class="leading-6 rounded-sm inline-flex text-center items-center justify-center align-middle cursor-pointer whitespace-no-wrap ml-3 bg-transparent border-none p-0 h-auto border-0 text-dark-2"
```
larsdouweschuitema, line 33: Achievement unlocked: LONG CLASS NAME 🥇

I guess this is the downside of not using one class and assigning styles, but trying to style everything using ui classes.

jpsc (approved):

src/components/ui-notification.vue
dine: Using the @ alias here will break when we import this components into another project (that has different path for @).

bartderijk (dismissed):

PR #142, opened 2020-11-12
Radio component

renancarvalho:
Add Radio button

I did the 'complicated' selectors using pure css. I didn't not figured how to do that with tailwind. Maybe someone could help on that, although I think this is something also to be discussed once it will add some complexity for something simple.

Let me know what you think!


src/components/ui-radio.vue
```diff
@@ -0,0 +1,61 @@
...
+                name="funda_radio_input"
+                class="hidden"
+                type="radio"
+                v-bind="$attrs"
+            />
+            <ui-label
+                class="cursor-pointer flex py-1 hover:text-blue-1"
+                :for="item.value"
+            >
+                <img
```
dine, line 15: Can we use a div here? The img tag creates a weird shape (not super round) in all browsers.

src/components/ui-radio.vue
```diff
@@ -0,0 +1,61 @@
+<template>
+    <div>
+        <div v-for="item in items" :key="item.value" class="py-2">
+            <input
+                :id="item.value"
+                name="funda_radio_input"
```
larsdouweschuitema, line 6: For accessibility reasons you need to have a dynamic name property instead of a static string that aligns with the label.

src/components/ui-radio.vue
```diff
@@ -0,0 +1,61 @@
...
+input[type='radio']:checked ~ label > img {
+    background-image: url(data:image/svg+xml;base64,PHN2ZyB4bWxucz0iaHR0cDovL3d3dy53My5vcmcvMjAwMC9zdmciIHZpZXdCb3g9IjAgMCAxNiAxNiI+CiAgPGNpcmNsZSBjeD0iOCIgY3k9IjgiIHI9IjQiIGZpbGw9IiMwMDUxN2YiLz4KPC9zdmc+Cg==);
+}
+input[type='radio']:hover ~ label > img {
+    background-image: url(data:image/svg+xml;base64,PHN2ZyB4bWxucz0iaHR0cDovL3d3dy53My5vcmcvMjAwMC9zdmciIHZpZXdCb3g9IjAgMCAxNiAxNiI+CiAgPGNpcmNsZSBjeD0iOCIgY3k9IjgiIHI9IjQiIGZpbGw9IiNjMWRiZTYiLz4KPC9zdmc+Cg==);
+}
+input[type='radio']:checked:hover ~ label > img {
+    background-image: url(data:image/svg+xml;base64,PHN2ZyB4bWxucz0iaHR0cDovL3d3dy53My5vcmcvMjAwMC9zdmciIHZpZXdCb3g9IjAgMCAxNiAxNiI+CiAgPGNpcmNsZSBjeD0iOCIgY3k9IjgiIHI9IjQiIGZpbGw9IiMwMDUxN2YiLz4KPC9zdmc+Cg==);
+}
+input[type='radio']:disabled ~ label > img {
```
larsdouweschuitema, line 51: Maybe you think of a cleaner way to store the data value of the url

src/components/ui-radio.vue
```diff
@@ -0,0 +1,61 @@
+<template>
+    <div>
```
larsdouweschuitema, line 2: By wrapping the input in a label you can avoid having to use an extra HTML tag. You can still get the same result visually.

src/components/ui-radio.vue
```diff
@@ -0,0 +1,61 @@
+<template>
```
larsdouweschuitema, line 1: Instead of having a ui-radio, we should have a ui-input that decides dynamically whether you want to use a radio or a checkbox, so that we avoid having to create two components with similar semantics, styling and behavior.

larsdouweschuitema (changes requested):
Nice that you started to get involved in the library. Good job Sir 👍 💯

src/components/ui-radio.vue
bartderijk: I'm a bit torn, on one hand I agree the radio and checkbox have a lot in common and on the other hand I feel they have such specific properties that they deserve to be a component on their own. For the sake of simplicity and avoid creating a jack of all kind trades component I'd argue having them separatly for now.

On the second point: I'd like to move forward in the most pragmatic manner possible, getting basic components out there rapidly so we can have a conversation based on what we have instead of what might be. In a past similar project, the team spent over a year theorizing with UX what it should be. In my experience that conversation can go a lot faster if there is a base set of components to begin with, acknowledging that they need more finetuning.

src/components/ui-radio.vue
renancarvalho: Indeed Dine!

src/components/ui-radio.vue
renancarvalho: This is how I make them a radio button group( only one can be selected ). 

could you explain more the relation of it with accessibility?

src/components/ui-radio.vue
renancarvalho: It is a direction too, I don't think one is better than the other... just different.
I would rather have it in two components, for simplicity.

src/components/ui-radio.vue
renancarvalho: Using variables.

src/components/ui-radio.vue
renancarvalho: Cool

src/components/ui-radio.vue
```diff
@@ -0,0 +1,66 @@
...
+        },
+        selected: {
+            type: String,
+        },
+    },
+};
+</script>
+<style scoped>
+label {
+    --input_checked: url(data:image/svg+xml;base64,PHN2ZyB4bWxucz0iaHR0cDovL3d3dy53My5vcmcvMjAwMC9zdmciIHZpZXdCb3g9IjAgMCAxNiAxNiI+CiAgPGNpcmNsZSBjeD0iOCIgY3k9IjgiIHI9IjQiIGZpbGw9IiMwMDUxN2YiLz4KPC9zdmc+Cg==);
```
dine, line 43: yay no IE11 support 🎉

dine (approved):

src/components/ui-radio.vue
renancarvalho: Dine 2020 Almost 2021.. what is IE ?

dine (approved):

renancarvalho:
Can we move it forward?

src/stories/ui-radio.stories.js
```diff
@@ -37,4 +37,7 @@ Input.args = {
     ],
     selected: 'option1',
     disabled: false,
+    onChange() {
+        console.log('renan');
```
sukalas-a, line 41: onChange log renan 🥇

renancarvalho: removing =)

dine (dismissed):

src/components/ui-radio.vue
renancarvalho: @larsdouweschuitema  i think this comment is blocking the merge.

src/components/ui-radio.vue
```diff
@@ -0,0 +1,75 @@
...
+                <div
+                    class="bg-blue-5 mr-2 mt-1 w-4 h-4 border border-blue-2 rounded-3xl"
+                ></div>
+                {{ item.label }}
+            </ui-label>
+        </div>
+    </div>
+</template>
+<script>
+import UiLabel from '@/components/ui-label.vue';
```
dine, line 25: We need to import using the relative path `from './ui-label.vue'`

renancarvalho: Done!

sukalas-a (approved):

PR #143, opened 2020-11-12
Spinner component

bartderijk:


src/components/ui-spinner.vue
```diff
@@ -0,0 +1,24 @@
+<template>
+    <div
+        v-if="show"
+        class="spinner animate-spin absolute w-6 h-6 box-border border-3 border-blue-2 rounded-full -mt-3 -ml-3 inset-1/2"
```
larsdouweschuitema, line 4: This "spinner" class seems vulnerable to be overwritten. Suggestion: component name ui-spinner

src/components/ui-spinner.vue
```diff
@@ -0,0 +1,24 @@
...
+        show: {
+            type: Boolean,
+            default: false,
+        },
+    },
+};
+</script>
+
+<style scoped>
+.spinner {
```
larsdouweschuitema, line 20: Can these properties not be done using tailwind?

tailwind.config.js
```diff
@@ -25,6 +25,12 @@ module.exports = extendConfig({
                 'fade-right-leave-active':
                     'fd-fade-right 0.6s 1 ease-out reverse',
             },
+            borderWidth: {
+                '3': '3px',
+            },
+            inset: {
```
larsdouweschuitema, line 31: This property name feels confusing when you don't look it up. Can we think of a more descriptive name?

src/components/ui-spinner.vue
```diff
@@ -0,0 +1,24 @@
+<template>
```
larsdouweschuitema, line 1: Just a general suggestion. Maybe you can think about what types of spinners we are using cross project and how we can determine the type of loading spinner we want to use. 
I would suggest aligning with UX on what spinner patterns we have and what types of properties you need to add, for example:

- Are we using different spinner patterns so that we need to add a property like spinner-type
- Colours
- Sizes

I am not saying you have to implement that all directly, but it might be good having thought about this and seeing what you can be implement versus what we want to implement along the way.

src/components/ui-spinner.vue
larsdouweschuitema, line 4: I would suggest using dynamic classes for certain classes thinking about how we can easily alter the spinner based on the UX patterns we have. For example: maybe I want an orange spinner, maybe I want to change the size of the spinner.

larsdouweschuitema (changes requested):
Added some comments that require some attention, in general: thanks for the good work on this and using tailwind 👍 💯

src/components/ui-spinner.vue
bartderijk: Single sided borders cannot, unless we use a plugin but that felt like overhead. Also for content I couldn't find anything.

tailwind.config.js
bartderijk: Inset is what tailwind uses for this, it takes a little getting used to but I think it's not a good idea to deviate from that standard https://tailwindcss.com/docs/top-right-bottom-left#app

src/components/ui-spinner.vue
bartderijk: Agreed! I went for the most minimalistic version to start building with, using this as a base we can create more elaborate patterns.

src/components/ui-spinner.vue
bartderijk: That's an interesting idea, but for now I'd like to restrict it a bit to prevent a multitude of rainbow colored spinners. I'll introduce size, let's make the component adapt to our needs as we go.

bartderijk: The color is actually something we would indeed need, given it being used on various colored backgrounds. I'm adding some basic options.

tailwind.config.js
larsdouweschuitema: I did not know that, It thought it was custom. Never mind this comment then.

larsdouweschuitema (approved):

src/components/ui-spinner.vue
```diff
@@ -0,0 +1,54 @@
...
+            }[this.tone];
+            return classes;
+        },
+    },
+};
+</script>
+
+<style scoped>
+.ui-spinner {
+    content: '';
```
dine, line 51: We don't need this

tailwind.config.js
```diff
@@ -25,6 +25,12 @@ module.exports = extendConfig({
                 'fade-right-leave-active':
                     'fd-fade-right 0.6s 1 ease-out reverse',
             },
+            borderWidth: {
```
dine, line 28: By adding this config here we removed all other border sizes from default tailwind (like `border-2`, `border-4`). Maybe let's use a spinner with 2px border width? Or maybe try the divideWidth: https://tailwindcss.com/docs/divide-width#customizing

src/components/ui-spinner.vue
```diff
@@ -0,0 +1,54 @@
...
+    border-right-color: transparent;
```
dine, line 52: https://github.com/tailwindlabs/tailwindcss/pull/560#issuecomment-670045304

PR #146, opened 2020-11-13
Data list

bartderijk:
Added the basics, looking for feedback.


src/stories/ui-data-list.stories.js
```diff
@@ -0,0 +1,25 @@
...
+
+const Template = (args, { argTypes }) => ({
+    props: Object.keys(argTypes),
+    components: { UiDataList },
+    template: '<ui-data-list :card="card" :data="data"></ui-data-list>',
+});
+
+export const DataList = Template.bind({});
+DataList.args = {
+    data: [{ Name: 'Jane Doe' }, { Phone: '020 561 8200' }],
```
dine, line 23: What do you think instead of sending an object we create `<ui-dl>`, `<ui-dt>` and `<ui-dd>` so this way we can use slots and maybe send different things inside? For example, inside the data-list we could use `<ui-icon>` or `<ui-input>`

dine: I'm ok with using like this if we don't feel the need composition for this component. This is a question for everyone @funda-frontend/devs ☝️

renancarvalho: I think it is a great idea @dine, it will make this guy more adaptable.
@bartderijk  let me know if you need help on this.

bartderijk: I like it! All in favor? Speak now or forever hold your peace :)

bartderijk: Done!

dine (approved):

src/components/ui-dd.vue
```diff
@@ -0,0 +1,19 @@
+<template>
```
larsdouweschuitema, line 1: I would rename these components dd and dt to "data-description" and "data-term" because that is what they are, and it makes it more descriptive.

**Definition and Usage**
The <dd> tag is used to describe a term/name in a description list.
The <dd> tag is used in conjunction with <dl> (defines a description list) and <dt> (defines terms/names).

larsdouweschuitema (changes requested):

src/components/ui-dd.vue
bartderijk: Nice one!

larsdouweschuitema (approved):

PR #147, opened 2020-11-13
Add checkbox component

renancarvalho:
Checkbox for us <3

### DIFF
--- a/src/components/ui-checkbox.vue
+++ b/src/components/ui-checkbox.vue
@@ -1,0 +1,76 @@
+<template>
+    <div>
+        <div v-for="item in items" :key="item.value" class="py-2">
+            <input
+                :id="item.value"
+                name="funda_radio_input"
+                class="hidden"
+                type="checkbox"
+                v-bind="$attrs"
+                v-on="$listeners"
+            />
+            <ui-label
+                class="cursor-pointer flex py-1 hover:text-blue-1"
+                :for="item.value"
+            >
+                <div
+                    class="bg-blue-5 border rounded-sm mr-2 mt-1 w-4 h-4 border-blue-2"
+                ></div>
+                {{ item.label }}
+            </ui-label>
+        </div>
+    </div>
+</template>
+<script>
+import UiLabel from '@/components/ui-label.vue';
+
+export default {
+    components: {
+        UiLabel,
+    },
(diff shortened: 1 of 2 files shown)


src/components/ui-checkbox.vue
```diff
@@ -0,0 +1,76 @@
+<template>
+    <div>
```
bartderijk, line 2: We don't need this div

src/components/ui-checkbox.vue
```diff
@@ -0,0 +1,76 @@
...
+import UiLabel from '@/components/ui-label.vue';
+
+export default {
+    components: {
+        UiLabel,
+    },
+    inheritAttrs: false,
+    props: {
+        items: {
+            type: [Array, String],
```
bartderijk, line 34: What scenario do we need a string for?

src/stories/ui-checkbox.stories.js
```diff
@@ -0,0 +1,45 @@
...
+            value: 'option3',
+        },
+        {
+            label: 'Option 4',
+            value: 'option4',
+        },
+    ],
+    selected: 'option1',
+    disabled: false,
+    onChange() {},
```
bartderijk, line 44: Do we need the onChange?

bartderijk:
> Checkbox for us <3

Nice work, thanks for lending a hand over here!

src/components/ui-checkbox.vue
renancarvalho: Think we do once the next one is a for.. 

if I remove I get this from eslint - https://eslint.vuejs.org/rules/valid-template-root.html

image

src/stories/ui-checkbox.stories.js
renancarvalho: Was just to validate if it was/is being called. not necessary at all, but there just as an indicator..

src/components/ui-checkbox.vue
bartderijk: I see, thanks for clearifying!
 
Schermafbeelding 2020-09-11 om 13 13 51

renancarvalho:
Can we move it forward?

bartderijk (approved):

PR #159, opened 2020-11-19
Tooltip component

bartderijk:
Added tooltip component


jpsc:
@bartderijk I don't think this will be enough.
This needs to have some logic regarding tooltip position.

My best suggestion is to look on how to best integrate [popper.js](https://popper.js.org/).
There is a [vue version](https://github.com/RobinCK/vue-popper) or you can directly integrate.

bartderijk:
> @bartderijk I don't think this will be enough.
> This needs to have some logic regarding tooltip position.
> 
> My best suggestion is to look on how to best integrate [popper.js](https://popper.js.org/).
> There is a [vue version](https://github.com/RobinCK/vue-popper) or you can directly integrate.

Nice one, I've added popper so we can get proper positioning. It needs some more finetuning but it's already a nice improvement.

tailwind.config.js
```diff
@@ -46,6 +49,9 @@ module.exports = {
                 '1/3': '33%',
                 '2/3': '66%',
             },
+            translate: {
+                '-1/2': '-50%',
```
jpsc, line 53: I think this already exists on tailwind default config.

.-translate-x-1/2 | --transform-translate-x: -50%;
.-translate-y-1/2 | --transform-translate-y: -50%;



-- | --

bartderijk: Nice one, tho translate and left can be removed all together anyways on this one, popper will be taking care of the caret.

jpsc: Sure. Just remember that every custom class we add here we need to add it to the funda tailwind config.

src/components/ui-tooltip.vue
```diff
@@ -0,0 +1,61 @@
+<template>
+    <popper :key="popperKey" :trigger="trigger" :options="{ placement: side }">
```
bartderijk, line 2: Key is necessary to make Popper re-render correctly:
https://michaelnthiessen.com/force-re-render/#key-changing-to-force-re-renders-of-a-component

Long story short: popper does not update it's props dynamically, it requires you to call its update function. Popper V2 has a nice way to do that while taking new options into account but the Vue adapter we're using is using Popper V1 which won't update with new options even if you beat it with a digital stick (using this.$refs to hack directly into it).

So forcing it to re-render with the key was the cleanest way to go.

src/components/ui-tooltip.vue
```diff
@@ -0,0 +1,61 @@
...
+        <div
```
bartderijk, line 3: Todo: Customise propper's styling to conform to funda ui

.storybook/webpack.config.js
```diff
@@ -5,7 +5,15 @@ module.exports = ({config}) => {
         test: /\.css$/,
         use: ['postcss-loader' ],
         include: path.resolve(__dirname, "../")
-    });
+    },
+    {
+        test: /\.postcss$/,
```
jpsc, line 10: Adding this and the post postcss style tag will break if used on the website.
This will also create a new requirement to use UI (wherever we want to).

bartderijk: How so? Postcss is already part of the project, and Tailwind recommends using it  as the preprocessor https://tailwindcss.com/docs/using-with-preprocessors#using-post-css-as-your-preprocessor

src/styles/popper.css
```diff
@@ -0,0 +1,53 @@
+/* Cusomised Popper styles, do not overwrite */
+
```
bartderijk, line 2: Added the popper styles as an external css file because the caret (arrow triangle in the tooltip) is dynamically added and thus would not receive any scoped styles coming from the tooltip component. Removing the scoped attribute means the styles are applied globally, so it made more sense to me to make the tooltips use this shared stylesheet as opposed to having it in each and every single tooltip component.

bartderijk:
Hi all, Can we move this one forward? It's a fully fledged tooltip, all dressed up and ready to go :) I'd like to hear what you think!

src/styles/popper.css
dine: Screenshot 2020-11-24 at 10 55 28
The triangle is not getting the other background colors

src/styles/popper.css
```diff
@@ -0,0 +1,53 @@
...
+    border-style: solid;
+    position: absolute;
+    margin: 8px;
+}
+.popper[x-placement^='top'] {
+    margin-bottom: 8px;
+}
+.popper[x-placement^='top'] .popper__arrow {
+    border-width: 8px 8px 0 8px;
+    border-color: #fafafa transparent transparent transparent;
```
dine, line 15: light-3 is #f7f7f7 https://github.com/funda-frontend/tailwind-config/blob/master/index.js#L41

src/styles/popper.css
```diff
@@ -0,0 +1,53 @@
+/* Cusomised Popper styles, do not overwrite */
+
+.popper .popper__arrow {
```
dine, line 3: I think the this styles work better for showing the shadow https://git.funda.nl/projects/FRN/repos/ui/browse/src/components/_tooltip.scss#53-71

dine (commented):
Mixed feelings about adding custom styles on top of vue-popper 🤔

src/styles/popper.css
bartderijk: @dine  The comment is confusing I guess, it's not on top of other popper styles. This is all the styles there is, except for the inline positioning specific styles.

src/styles/popper.css
bartderijk: I dont know what happened, my fix for this wasn't pushed. I'll add it now.

bartderijk: Nevermind, my fix was less pretty. Let's go with your suggested changes.

bartderijk:
> Mixed feelings about adding custom styles on top of vue-popper 🤔

It's not on top of Popper styles, this is the only styles there are.

dine (approved):

PR #186, opened 2020-11-25
fix(tooltip): move popper styles to components and include on npm package

jpsc:
When packaging for npm, styles from popper, were not being included.


src/components/ui-tooltip.vue
```diff
@@ -22,7 +22,7 @@
 
 <script>
 import Popper from 'vue-popperjs';
-import '../styles/popper.css';
+import './../styles/vendors/popper.css';
```
bartderijk, line 25: The styles look at this point in very little ways like the one's supplied by the vendor. When we add it to the vendor folder we are implying that they are the vendor's supplied styles, I'm afraid it might be confusing.. don't you think?

bartderijk: For reference, here's the original: https://github.com/RobinCK/vue-popper/blob/master/dist/vue-popper.css and here's ours: https://github.com/funda-frontend/ui/blob/7cc7083bb99756660b0a1b6bcdf8acc19236d614/src/styles/vendors/popper.css

dine (approved):

src/components/ui-tooltip.vue
jpsc: Sure, i just wanted a directory inside of styles so I could exclude storybook styles from being included on the package.
I changed it to components.

dine (approved):

bartderijk (approved):

PR #187, opened 2020-11-26
Button group

bartderijk:
I added the button group component, it can take some buttons and will group them together just like it does in the funda ui button group component. Nothing fancy (yet), it just does exactly the same job as it's predecessor. It's very minimal and basic, a good start for iteration I figure. Let me know what you think.

Schermafbeelding 2020-11-26 om 13 48 03

### DIFF
--- a/src/components/ui-button-group.vue
+++ b/src/components/ui-button-group.vue
@@ -1,0 +1,16 @@
+<template>
+    <div
+        class="fd-btn-group inline-flex"
+        role="group"
+        aria-label="Button group"
+    >
+        <slot />
+    </div>
+</template>
+
+<style scoped>
+.fd-btn-group .fd-btn:not(:last-child) {
+    border-top-right-radius: 0;
+    border-bottom-right-radius: 0;
+}
+</style>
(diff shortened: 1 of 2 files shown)


src/components/ui-button-group.vue
```diff
@@ -0,0 +1,24 @@
...
+        class="fd-btn-group inline-flex"
+        role="group"
+        aria-label="Button group"
+    >
+        <slot />
+    </div>
+</template>
+
+<style scoped>
+.fd-btn-group button:not(:last-child),
```
bartderijk, line 12: I attempted to have these styles applied using tailwind, but no sigar. I have not yet found a way to do it using their utility classes.

dine: I really don't like this styles here, specially the `:not` selector. 
Maybe something using only :first-child and :last-child makes it easier to read?

bartderijk: I agree, tho the styles are an exact copy of what we're already using on funda ui.

bartderijk: @dine I refactered it so it's a bit cleaner.

dine (approved):

PR #212, opened 2020-12-14
feat(input-id): added support for radio groups using unique ids

sukalas-a:
Added this improvements as I was having issues with the value bindings in Flex-ui. 
Also added a name prop to add radio groups support

### DIFF
--- a/src/components/ui-radio.vue
+++ b/src/components/ui-radio.vue
@@ -1,8 +1,8 @@
 <template>
     <div>
-        <div v-for="item in items" :key="item.value" class="py-2">
+        <div v-for="(item, index) in items" :key="item.value" class="py-2">
             <input
-                :id="(item.label + '_' + id) | trimEmptySpace"
+                :id="UUIDs[index]"
                 :name="name"
                 class="hidden"
                 type="radio"
@@ -11,7 +11,7 @@
             />
             <ui-label
                 class="cursor-pointer flex py-1 hover:text-blue-1"
-                :for="(item.label + '_' + id) | trimEmptySpace"
+                :for="UUIDs[index]"
             >
                 <div
                     class="bg-blue-5 mr-2 mt-1 w-4 h-4 border border-blue-2 rounded-3xl"
@@ -26,12 +26,6 @@ import UiLabel from './ui-label.vue';
 export default {
     components: {
         UiLabel,
-    },
-    filters: {
-        trimEmptySpace(str) {
-            if (!str) return;
-            return str.replace(/\s/g, '');
-        },
     },
     inheritAttrs: false,
     props: {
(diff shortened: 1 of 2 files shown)


src/components/ui-radio.vue
```diff
@@ -2,16 +2,16 @@
     <div>
         <div v-for="item in items" :key="item.value" class="py-2">
             <input
-                :id="item.value"
-                name="funda_radio_input"
+                :id="(item.label + '_' + id) | trimEmptySpace"
```
larsdouweschuitema, line 5: Maybe you can move this to a method that return this as a variable. Also to avoid code duplication for id + for.
Besides that all good 👍

larsdouweschuitema (changes requested):

src/components/ui-radio.vue
```diff
@@ -27,16 +27,35 @@ export default {
...
+            default: '',
         },
     },
+    data() {
+        return {
+            id: null,
+        };
+    },
+    mounted() {
+        this.id = this._uid;
```
TobbenTM, line 57: `this._uid` is not a public API, and more importantly, doesn't work well with snapshot tests. Is this really the best way?

sukalas-a: Probably not the best but the simplest.
As far as the testing, It can be mocked right?
As far as the private API point fair enough I guess.
I wanted to avoid making a method that does pseudo UIDs but maybe it's indeed a more solid approach.

TobbenTM: I think we tried mocking it for some tests, but then we ran into other issues. If you're curious, I think Alexandru should know more.

src/components/ui-radio.vue
```diff
@@ -33,8 +33,40 @@ export default {
...
+            type: String,
+            default: 'funda_radio_input',
+        },
         selected: {
             type: String,
+            default: '',
+        },
+    },
+    computed: {
+        UUIDs() {
```
dine, line 46: I'd suggest a different name, if we would really need 'uuid'here we could use a package https://en.wikipedia.org/wiki/Universally_unique_identifier

src/components/ui-radio.vue
```diff
@@ -33,8 +33,40 @@ export default {
...
+            this.items.forEach((item) => {
+                array.push(
+                    this.shuffle(item.label).join('').replace(/\s/g, '')
+                );
+            });
+            return array;
+        },
+    },
+    methods: {
+        shuffle(string) {
```
dine, line 57: do we need all this?

sukalas-a: I refactored like we discussed. 
To sum up, we agreed that in order to keep the components simple we better let the parent handle the id complexity.

dine (approved):

PR #225, opened 2020-12-22
fix(classnames): change prefix classnames to avoid clashing

dine:


daniarsuprianto (approved):

src/components/ui-button-group.vue
```diff
@@ -9,20 +9,20 @@
 </template>
 
 <style scoped>
-.fd-btn-group button,
-.fd-btn-group a {
```
sukalas-a, line 13: I guess this a temporary fix as we shouldn't be needing these styles here when we can style Tailwind across everywhere.

Another approach could be to remove these classes (only the ones for the slot elements) from here and add them where these components are used. So when you pass the html to the slot you also pass the classes needed and then refactor all together with ctr + f

sukalas-a (approved):

PR #227, opened 2020-12-22
feat(checkbox): seperated checkbox input from group

sukalas-a:
Separated the checkbox input from the group for more flexibility

### DIFF
--- a/src/components/ui-checkbox-group.vue
+++ b/src/components/ui-checkbox-group.vue
@@ -1,6 +1,6 @@
 <template>
     <div>
-        <UiCheckbox
+        <ui-checkbox
             v-for="item in items"
             :id="`${id}_${item.value}`"
             :key="item.label"
@@ -11,7 +11,7 @@
             @change="$emit('change', $event.target.value)"
         >
             {{ item.label }}
-        </UiCheckbox>
+        </ui-checkbox>
     </div>
 </template>
 <script>
@@ -21,7 +21,6 @@ export default {
     components: {
         UiCheckbox,
     },
-    inheritAttrs: false,
     props: {
         items: {
             type: Array,
@@ -34,10 +33,6 @@ export default {
         name: {
             type: String,
             default: 'funda_checkbox_input',
-        },
-        selected: {
-            type: [String, Boolean, Number],
-            default: '',
         },
         disabled: {
             type: Boolean,
--- a/src/components/ui-checkbox-group.vue
+++ b/src/components/ui-checkbox-group.vue
@@ -41,37 +41,3 @@ export default {
     },
 };
 </script>
-<style scoped>
-label {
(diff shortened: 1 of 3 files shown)


src/components/ui-checkbox-group.vue
```diff
@@ -0,0 +1,70 @@
...
+            default: '',
+        },
+        disabled: {
+            type: Boolean,
+            default: false,
+        },
+    },
+};
+</script>
+<style scoped>
```
dine, line 37: This classes should be part of `ui-checkbox.vue` file

src/components/ui-checkbox-group.vue
```diff
@@ -0,0 +1,70 @@
...
+    </div>
+</template>
+<script>
+import UiCheckbox from './ui-checkbox.vue';
+
+export default {
+    components: {
+        UiCheckbox,
+    },
+    inheritAttrs: false,
```
dine, line 20: I think this can be removed

src/components/ui-checkbox-group.vue
```diff
@@ -0,0 +1,70 @@
+<template>
+    <div>
+        <UiCheckbox
```
dine, line 3: Are we using PascalCase or kebab-case?

src/stories/ui-checkbox-group.stories.js
```diff
@@ -0,0 +1,45 @@
...
+        {
+            label: 'Option 3',
+            value: 'option3',
+        },
+        {
+            label: 'Option 4',
+            value: 'option4',
+        },
+    ],
+    selected: 'option2',
```
dine, line 42: For multiple checkboxes we could send an Array with the selected values: https://vuejs.org/v2/guide/forms.html#Checkbox

src/components/ui-checkbox-group.vue
```diff
@@ -0,0 +1,70 @@
...
+        UiCheckbox,
+    },
+    inheritAttrs: false,
+    props: {
+        items: {
+            type: Array,
+            required: true,
+        },
+        selected: {
+            type: String,
```
dine, line 27: This could be an Array with the selected values

src/stories/ui-checkbox.stories.js
```diff
@@ -16,30 +16,12 @@ const Template = (args, { argTypes }) => ({
     props: Object.keys(argTypes),
     components: { UiCheckbox },
     template:
-        '<ui-checkbox :selected="selected" :items="items" @change="onChange" :disabled="disabled" />',
+        '<ui-checkbox :id="id" :disabled="disabled"> Option 1 </ui-checkbox>',
```
dine, line 19: Let's update the example sending :checked="true" (just to be sure the attributes are being inherited)

dine (commented):
I added some suggestions

src/components/ui-checkbox-group.vue
sukalas-a: Not sure what is the best way but for now I added an extra property selected inside the items array. Mostly because that's the structure we use in Flex-UI
If the structure is different in the Website we should probably refactor later.

dine (approved):

PR #250, opened 2021-01-18
Nuxt support

daniarsuprianto:
This PR will merge ui components and nuxt components (in separate nuxt app). This way it becomes tree-shakable and can be imported automatically
https://nuxtjs.org/blog/improve-your-developer-experience-with-nuxt-components/#third-party-component-library

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,5 +4,37 @@
 
 This project is currently a work in progress, you'll find some components but it's incomplete at the moment.
 
+# Usage
+
+Import in local component:
+
+```js
+// MyComponent.vue
+
+import UiButton from '@funda/ui/src/components/ui-button.vue';
+```
+
+or
+
+```js
+// MyComponent.vue
+
+import { UiButton } from '@funda/ui';
+```
+
+If you are using [Nuxt auto import component](https://nuxtjs.org/docs/2.x/directory-structure/components):
+
+```js
+// nuxt.config.js
+
+export default {
+  ...,
+  components: true,
+  buildModules: ["@funda/ui/src/nuxt.js"],
+  ...
+}
+```
+
 # Latest
+
 https://funda-frontend.github.io/ui/latest/
--- a/src/components/ui-icon.vue
+++ b/src/components/ui-icon.vue
@@ -15,7 +15,7 @@ export default {
     },
     computed: {
         thisIcon() {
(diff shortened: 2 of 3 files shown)


jpsc:
Couple questions:

Does this mean that we no longer need to create this file?

```javascript
// ~/plugins/funda-ui.js
import Vue from 'vue'
import { UiInput, UiButton, UiIcon } from '@funda/ui'

Vue.component('UiInput', UiInput)
Vue.component('UiButton', UiButton)
Vue.component('UiIcon', UiIcon)

```

Will this change work if you don't use `components: true` on `nuxt.config.js`? Because we have been disabling this for now.

daniarsuprianto:
@jpsc 
> Does this mean that we no longer need to create this file?
> 
> ```js
> // ~/plugins/funda-ui.js
> import Vue from 'vue'
> import { UiInput, UiButton, UiIcon } from '@funda/ui'
> 
> Vue.component('UiInput', UiInput)
> Vue.component('UiButton', UiButton)
> Vue.component('UiIcon', UiIcon)
> ```
Yes, this file is no longer needed, but need to set `components: true`.

> Will this change work if you don't use `components: true` on `nuxt.config.js`? Because we have been disabling this for now.

I think it will work, as long as we don't import this file into nuxt.config.js in the buildModules.
```js
buildModules: ["@funda/ui/src/nuxt.js"]
```
im curious why do we disable the `components: true`?

README.md
```diff
@@ -4,5 +4,37 @@
...
+```
+
+If you are using [Nuxt auto import component](https://nuxtjs.org/docs/2.x/directory-structure/components):
+
+```js
+// nuxt.config.js
+
+export default {
+  ...,
+  components: true,
```
dine, line 32: I think we don't need this extra file:
https://git.funda.nl/projects/WEBAPI/repos/makelaarsearch.funda.io/browse/nuxt.config.js#19

daniarsuprianto: you can just set it to true, yes. but this PR is up for discussion, if we want to do it like yours, or use this config

jpsc:
> @jpsc 
> im curious why do we disable the `components: true`?

Just because it's one of those things you need to know it exists. And since it was a new feature of Nuxt it would make easier to be explicit when using a component.

jpsc (approved):

PR #265, opened 2021-01-28
Ui-card media slot

bartderijk:
I've found a need for a media slot in the ui-card, this is a pattern seen commonly in funda desk already. For this I added a slot and added styling to have the media always filling out the top of the card.

### DIFF
--- a/src/components/ui-card-image.vue
+++ b/src/components/ui-card-image.vue
@@ -1,0 +1,22 @@
+<template>
+    <img
+        class="absolute block w-full h-full top-0 left-0 object-cover"
+        :src="src"
+        :alt="alt"
+    />
+</template>
+
+<script>
+export default {
+    props: {
+        src: {
+            required: true,
+            type: String,
+        },
+        alt: {
+            required: true,
+            type: String,
+        },
+    },
+};
+</script>
--- a/src/components/ui-card-image.vue
+++ b/src/components/ui-card-image.vue
@@ -1,13 +1,25 @@
 <template>
-    <img
-        class="absolute block w-full h-full top-0 left-0 object-cover"
-        :src="src"
-        :alt="alt"
-    />
+    <ui-card>
+        <template #image>
+            <div class="overflow-hidden relative h-0 pt-40 md:rounded-t-sm">
+                <img
+                    class="absolute block w-full h-full top-0 left-0 object-cover"
+                    :src="src"
+                    :alt="alt"
+                />
+            </div>
+        </template>
+        Card content
+    </ui-card>
 </template>
 
 <script>
+import UiCard from './ui-card.vue';
+
 export default {
+    components: {
+        UiCard,
+    },
     props: {
         src: {
             required: true,
--- a/src/components/ui-card.vue
+++ b/src/components/ui-card.vue
@@ -11,7 +11,7 @@
         </div>
         <div
             v-if="hasMedia"
-            class="mb-4 overflow-hidden relative h-0 pt-40 md:rounded-t-sm"
+            class="overflow-hidden relative h-0 pt-40 md:rounded-t-sm"
         >
             <slot name="media" />
         </div>
--- a/src/components/ui-card.vue
+++ b/src/components/ui-card.vue
@@ -11,7 +11,7 @@
         </div>
         <div
             v-if="hasMedia"
-            class="card__image mb-4 overflow-hidden relative h-0 pt-40 md:rounded-t-sm"
+            class="mb-4 overflow-hidden relative h-0 pt-40 md:rounded-t-sm"
         >
             <slot name="media" />
         </div>
--- a/src/components/ui-card.vue
+++ b/src/components/ui-card.vue
@@ -9,13 +9,13 @@
         <div v-if="hasNotifications" class="border-b border-light-2">
             <slot name="notifications" />
         </div>
+        <div
+            v-if="hasMedia"
+            class="card__image mb-4 overflow-hidden relative h-0 pt-40 md:rounded-t-sm"
+        >
+            <slot name="media" />
+        </div>
         <div class="py-4 px-4 md:px-6">
-            <div
-                v-if="hasMedia"
-                class="card__image mb-4 overflow-hidden relative h-0 pt-40"
-            >
-                <slot name="media" />
-            </div>
             <slot />
         </div>
     </div>
--- a/src/components/ui-card.vue
+++ b/src/components/ui-card.vue
@@ -10,6 +10,16 @@
             <slot name="notifications" />
         </div>
         <div class="py-4 px-4 md:px-6">
+            <div
+                v-if="hasMedia"
+                class="card__image mb-4 overflow-hidden relative h-0 pt-40"
+            >
+                <div
+                    class="absolute block w-full top-0 left-0 h-full object-cover"
+                >
+                    <slot name="media" />
+                </div>
+            </div>
             <slot />
         </div>
     </div>
@@ -23,6 +33,9 @@ export default {
         },
         hasNotifications() {
             return Boolean(this.$slots['notifications']);
+        },
+        hasMedia() {
+            return Boolean(this.$slots['media']);
         },
     },
 };
--- a/src/components/ui-card.vue
+++ b/src/components/ui-card.vue
@@ -9,12 +9,7 @@
         <div v-if="hasNotifications" class="border-b border-light-2">
             <slot name="notifications" />
         </div>
-        <div
-            v-if="hasMedia"
-            class="overflow-hidden relative h-0 pt-40 md:rounded-t-sm"
-        >
-            <slot name="media" />
-        </div>
+        <slot name="image" />
         <div class="py-4 px-4 md:px-6">
             <slot />
         </div>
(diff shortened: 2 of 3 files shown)


src/components/ui-card.vue
```diff
@@ -24,6 +30,21 @@ export default {
...
         },
+        hasMedia() {
+            return Boolean(this.$slots['media']);
+        },
     },
 };
 </script>
+
+<style scoped>
+.card__image * {
```
bartderijk, line 41: As slots do not accept classes, it is not possible to pass tailwind magic to the media element. We can either style it like this, or add something like https://github.com/benface/tailwindcss-children. The latter I found too much overhead, which is why I opted for adding regular 'ol styles.

src/components/ui-card.vue
```diff
@@ -10,6 +10,12 @@
             <slot name="notifications" />
         </div>
         <div class="py-4 px-4 md:px-6">
+            <div
+                v-if="hasMedia"
+                class="card__image mb-4 overflow-hidden relative h-0 pt-40"
```
bartderijk, line 15: 'card__image' is needed to circumvent Vue's limitation to adding classes to slots. Check out my comment below at the styles for more detaills

src/components/ui-card.vue
daniarsuprianto: i think the style should be owned by the content of the slot itself, not defined here. what i mean is, this `card__image *` selector already tells how (whatever element in) the slot should look. but that's not always the case (or is it?) now if i need to adjust my slot to have 50% width, i need to do rewrite of these styles. consider to just remove this

larsdouweschuitema: I agree with Daniar that we need to have the flexibility depending on the use case. I am not a 100% sure to what extend you need to have classes in place for consistency reasons versus how much flexibility you need.

bartderijk: But then it's kinda missing the point, it want it to be styled by default to nicely fill the entire wrapper automatically. Otherwise I might as well add it to the body, as there's no added value. To answer your question: Yes, it always needs to look like this.

bartderijk: I think we need to develop mainly for our actual current requirements. I'm against adding functionality for things we don't need (yet). If the need arises, then we can be pragmatic and adjust for the need at hand.

larsdouweschuitema: I understand your point @bartderijk 
As long as people have the option to not render this at all, it is fine that we can merge this. And then when someone has an actual use case we can adapt it.

jpsc: How about creating a component that extends the Card but has the styles that you need?

```vue
// ui-card.vue
<template>
	<div class="bg-white shadow md:rounded-sm ...">
		<slot/>
	</div>
</template>
```
```vue
// ui-media-card.vue
<template>
	<ui-card>
		<div class="absolute t-0 ...">
			<slot name="media">
		</div>
		<slot/>
	</ui-card>
</template>
```

```vue 
// using ui-media-card
<template>
	<ui-media-card>
		<template #media>
			<img src="https://placekitten.com/200/300" alt="A kitty cat" />
		</template>
		The rest of the content
	</ui-media-card>
</template>
```

fwiw, this is of course a simplified version.

jpsc:
```vue
// ui-card
<template>
	// it's questionable that you need multiple slots
	<slot name="media"></slot>
	<slot />
</template>
```

```vue
// ui-card-with-media-image
<template>
	<ui-card>
		<template #media>
			<div class="mb-4 overflow-hidden ...">
				<slot name="image"/>
			</div>
		</template>
		<template #default>
			<slot />
		</template
	</ui-card>
</template>
```

```vue
// ui-card-media-image
<template>
	<img v-bind="$props" class="absolute block w100 ...">
</template>
```

```vue
// using ui-card-with-media-image
<ui-card-with-media-image>
	<template #image>
		<ui-card-media-image :src="src" alt="alt" />
	</template>
	Content
</ui-card-with-media-image>

```

Too convuluted ?

bartderijk: Thanks for taking the time to come up with an example, the challenge here is that we want to support other types of media than img elements. We could create a component for every type of media, like ui-card-media-image, ui-card-media-video, ui-card-media-picture, etc. But indeed that becomes convoluted/bloated real quick.

src/components/ui-card.vue
```diff
@@ -10,6 +10,12 @@
             <slot name="notifications" />
         </div>
```
dine, line 11: I was expecting media not to have padding. We have that pattern at the fundadesk homepage for example: https://www.fundadesk.nl/start

Screenshot 2021-02-01 at 13 13 50

bartderijk: Excellent point, I'll work on that

bartderijk: - [ ] Media should not have padding.

bartderijk: fixed

src/components/ui-card.vue
```diff
@@ -9,6 +9,12 @@
         <div v-if="hasNotifications" class="border-b border-light-2">
             <slot name="notifications" />
         </div>
+        <div
+            v-if="hasMedia"
+            class="card__image mb-4 overflow-hidden relative h-0 pt-40 md:rounded-t-sm"
```
dine, line 14: We can remove `mb-4` here

src/components/ui-card.vue
dine: Maybe we should check if we need another tag type apart from `` right now? Maybe if we need we might not to do a cover-fit for example. 
I would like to see tailwind classes `absolute block w-full h-full top-0 left-0 object-cover` sent to the content somehow.

bartderijk: I'll add a video element, good point.

If we want to add the tailwind classes to the slot's content, the only solution I see is adding the plugin for that: 
https://github.com/benface/tailwindcss-children

src/components/ui-card.vue
bartderijk: > We can remove `mb-4` here

Fixed!

src/components/ui-card-image.vue
```diff
@@ -0,0 +1,22 @@
+<template>
```
daniarsuprianto, line 1: i was thinking in this component we could do something like
```
<ui-card>
    <template v-slot:media>
       <img class="absolute block w-full h-full top-0 left-0 object-cover"
           :src="src"
           :alt="alt"
       />
    </template>
    Card content
</ui-card>
```

so, when you need to use it, you can just do 
```
<ui-card-image src="https://placekitten.com/200/300" alt="A kitty cat" />
```
instead of
```
 <ui-card>
   <template v-slot:media>
      <ui-card-image src="https://placekitten.com/200/300" alt="A kitty cat" /> 
   </template>
   Card content
</ui-card>
```
this way you only need to import one component as well, which is `ui-card-image`.
but let me know what you think :)

bartderijk: I rather extend ui-card, to prevent code duplication. I like the flexibility and ease of maintenance personally more than having to import 2 components.

daniarsuprianto: exactly, the purpose is to prevent importing 2 components. if you do
```
<ui-card>
   <template v-slot:media>
      <ui-card-image src="https://placekitten.com/200/300" alt="A kitty cat" /> 
   </template>
   Card content
</ui-card>
```
then in my home.vue, i need to import UiCard and UiCardImage.

bartderijk: Ah I misunderstood. Yeah that could work.

src/components/ui-card-image.vue
```diff
@@ -0,0 +1,22 @@
...
+    <img
+        class="absolute block w-full h-full top-0 left-0 object-cover"
+        :src="src"
```
dine, line 4: I think we can remove this, since attributes are inherited by default to the root element

src/components/ui-card-image.vue
```diff
@@ -0,0 +1,22 @@
...
+        :alt="alt"
+    />
+</template>
+
+<script>
```
dine, line 9: I think we can remove props here

src/components/ui-card.vue
dine, line 41: We can remove this styles too

src/components/ui-card-image.vue
dine: I think if we just add an image to <ui-card> will add it inside the padded content

bartderijk: fixed

src/components/ui-card-image.vue
bartderijk: Can you elaborate? I'd like to have them required and type checked, what are you suggesting exactly?

dine: This comment made sense before, when the `ui-card-image` was only the `` tag. Then all attributes would be inherited by default to the root element. I thought it could be useful want to use `srcset` for example or [any other img attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#attributes).

Maybe we could still do that by [disabling the attribute inheritance and binding it to the img](https://vuejs.org/v2/guide/components-props.html#Disabling-Attribute-Inheritance)?

dine (approved):

src/stories/ui-card.stories.js
```diff
@@ -92,3 +93,11 @@ const TemplateTitleNotification = () => ({
                 </ui-card>`,
 });
 export const CardTitleNotification = TemplateTitleNotification.bind({});
+
+const TemplateImage = () => ({
```
daniarsuprianto, line 97: <3

src/components/ui-card-image.vue
bartderijk: We settled on the idea of a simple img component, I'd like to stick to the basics for now and add functionality like that soon as we need it. But it's a great suggestion, I'll keep it in mind for if we ever find the need to expand this puppy.

daniarsuprianto (approved):

PR #268, opened 2021-02-01
feat(select): Bind disabled and hidden for options

TobbenTM:
We would like to use placeholder values in selects, and so we need to bind the disabled state for options

### DIFF
--- a/src/components/ui-select.vue
+++ b/src/components/ui-select.vue
@@ -16,6 +16,8 @@
                 :key="option.value"
                 :value="option.value"
                 :selected="option.value === selected"
+                :disabled="option.disabled"
+                :hidden="option.hidden"
             >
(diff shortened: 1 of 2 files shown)


src/components/ui-select.vue
```diff
@@ -16,6 +16,8 @@
                 :key="option.value"
                 :value="option.value"
                 :selected="option.value === selected"
+                :disabled="option.disabled"
+                :hidden="option.hidden"
```
bartderijk, line 20: The `hidden` attribute is not supported by IE 6-10. Consider adding a class `hidden` to ensure it's invisibility.

dine: I think we only support IE11+ and soon not even that one :)

dine (approved):

PR #358, opened 2021-04-06
Accordion

daniarsuprianto:


src/components/ui-accordion.vue
```diff
@@ -0,0 +1,57 @@
...
+            this.updateValue(val);
+        },
+    },
+    mounted() {
+        this.isExpanded = this.open;
+    },
+    methods: {
+        updateValue(value) {
+            this.isExpanded = value;
+            this.$emit('change', value);
```
dine, line 53: do we need this?

daniarsuprianto: I bumped into a case where there's a chart inside the accordion. I don't want to load the chart on page load, it should only be loaded when user open the accordion. So here I emit an event when user open/close the accordion, and listen to that event to load the chart. Does it make sense or any other way of doing it?

dine: Not sure if there's another way of doing it, but I get it now and I think it makes sense. Thanks!

dine (approved):

PR #371, opened 2021-04-09
fix(button): custom disabled classes per tone

dine:


src/components/ui-button.vue
```diff
@@ -35,32 +35,42 @@ export default {
...
+                danger: `bg-red-1 border-red-1 hover:bg-red-3 hover:border-red-3 shadow ${withPadding} ${whiteText}`,
+            }[this.tone];
+            const disabledClasses = {
+                primary: `shadow ${withPadding} ${disabledBackground} ${whiteText}`,
+                'primary-alt': `shadow ${withPadding} ${disabledBackground} ${whiteText}`,
+                secondary: `shadow bg-white ${withPadding} ${disabledBorder}`,
+                'secondary-alt': `shadow ${withPadding} ${disabledBorder}`,
+                tertiary: 'text-light-1',
+                transparent: `${withPadding} ${disabledBorder}`,
+                danger: `shadow ${withPadding} ${disabledBackground} ${whiteText}`,
```
daniarsuprianto, line 63: wow... i wouldnt want to touch this button component ever again

daniarsuprianto: just improvement suggestion, but it's gonna be big breaking change maybe. now this component seems complex and has too many variations. is it an idea to actually split this component into subcomponents, per tone? something like <ui-button-primary>, etc. it extends <ui-button> component

daniarsuprianto (approved):

PR #376, opened 2021-04-12
fix(toaster): avoid purge animation classes

dine:


src/components/ui-toaster.vue
```diff
@@ -1,6 +1,10 @@
 <template>
     <div class="fixed top-0 right-0 overflow-x-hidden z-50 max-w-sm">
-        <transition-group tag="div" name="animate-fade-right">
+        <transition-group
+            tag="div"
+            enter-active-class="animate-fade-right-enter-active"
+            leave-active-class="animate-fade-right-leave-active"
+        >
```
daniarsuprianto, line 7: ugh

daniarsuprianto: we need to do this for every animation i guess? i read something like this, maybe we could implement that? i mean, to whitelist the vue default animation classes

daniarsuprianto: https://github.com/ky-is/vue-cli-plugin-tailwind/blob/master/generator/templates/postcss.config.js

daniarsuprianto (approved):

bartderijk (approved):

src/components/ui-toaster.vue
dine: yeah, the only catch of that solution is that if we have a lot of animations they won't be purged at all

PR #390, opened 2021-04-26
fix(card): allow card content to have custom classes

dine:


src/components/ui-card.vue
```diff
@@ -1,23 +1,30 @@
 <template>
     <div class="bg-white shadow md:rounded-sm">
-        <h2
-            v-if="hasTitle"
-            class="m-0 text-base py-4 px-4 md:px-6 border-b border-light-2"
-        >
+        <div v-if="hasTitle" :class="customTitleClass">
```
daniarsuprianto, line 3: why the `customTitleClass` has to be here? i mean, it can also be attached to the content of the slot. like
```
<template v-slot:title>
<h2 class="customTitleClass">I want sushi</h2>
</template>
```
Same as the customCardClass

dine: For the cases where we want to override the default classes (the default ones are set on the props default value)

daniarsuprianto: just my personal preference, i don't like passing custom class props to a component (unless it's nunjucks template HAHA), especially when we're already using slot, we can configure the classes from the slot itself, instead of making the core component 'custom'.

arjanvanrees (approved):

src/components/ui-card.vue
```diff
@@ -10,14 +10,20 @@
             <slot name="notifications" />
         </div>
         <slot name="image" />
-        <div class="py-4 px-4 md:px-6">
+        <div :class="customCardClass">
```
bartderijk, line 13: My concern would be that this opens the gates for a multitude of variations. What exactly is the behaviour that you would like to add? Cant we add that as a default option? E.G: `isHuge` or `isFixed`.

src/components/ui-card.vue
bartderijk: You can't add styles to the slot anymore, it has to be on the parent https://vuejs.org/v2/guide/migration.html#slot-Attribute-Styling-removed

src/components/ui-card.vue
larsdouweschuitema: I am also missing a bit of context. Please add it to the description, so that we can give a better judgment / more effective code review.

dine: I'll not merge it, that will make us add some negative margins for now. I can show some use cases on our next meeting.

PR #409, opened 2021-05-18
Feature/button improvements

larsdouweschuitema:
### Changes

-  Remove height property. The height should be consistent for all buttons, any deviations can be created as custom buttons on the applications themselves. This is also easier because we do not have renderless components because of the design system
- Add a small property so that we can support the two patterns we have for the paddings (px-2 and px-6)
- Add button test for all properties and click listener


renancarvalho (dismissed):
Looks good.

Nice we are writing unit testing <3

bartderijk (approved):

PR #418, opened 2021-05-20
test(select): add initial select test

larsdouweschuitema:
Initial test for select component

### DIFF
--- a/src/__tests__/ui-select.test.js
+++ b/src/__tests__/ui-select.test.js
@@ -3,6 +3,8 @@ import UiSelect from '@/components/ui-select.vue';
 
 describe('UiSelect', () => {
     let wrapper;
+    let select;
+    let options;
     beforeEach(() => {
         wrapper = mount(UiSelect, {
             propsData: {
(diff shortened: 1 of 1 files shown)


src/__tests__/ui-select.test.js
```diff
@@ -0,0 +1,78 @@
+import { mount } from '@vue/test-utils';
+import UiSelect from '@/components/ui-select.vue';
+
+describe('UiModal', () => {
+    let wrapper;
+    beforeEach(() => {
+        wrapper = mount(UiSelect, {
+            propsData: {
+                options: [
```
renancarvalho, line 9: I would make it a const variable on top.

src/__tests__/ui-select.test.js
```diff
@@ -0,0 +1,78 @@
...
+                        value: 'test3',
+                        hidden: true,
+                    },
+                ],
+            },
+        });
+    });
+
+    it('Should set selected option', async () => {
+        const select = wrapper.find('select');
```
renancarvalho, line 36: Maybe you could make it a global variable and only do it once on beforeEach

src/__tests__/ui-select.test.js
larsdouweschuitema: Moved the array to top of file

Zishrodrigues (approved):

PR #419, opened 2021-05-21
test(input): add unit test for ui-input

larsdouweschuitema:
Add initial unit test for ui-input and check prop on ui-select

### DIFF
--- a/src/__tests__/ui-input.test.js
+++ b/src/__tests__/ui-input.test.js
@@ -38,6 +38,9 @@ describe('UiInput', () => {
 
     it('Set slots', () => {
         expect(wrapper.text()).toContain(SLOT_PREFIX);
+        expect(input.classes()).toContain('pl-12');
+
         expect(wrapper.text()).toContain(SLOT_SUFIX);
+        expect(input.classes()).toContain('pr-12');
     });
 });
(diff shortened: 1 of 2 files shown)


src/__tests__/ui-input.test.js
```diff
@@ -0,0 +1,46 @@
...
+            slots: {
+                prefix: SLOT_PREFIX,
+                sufix: SLOT_SUFIX,
+            },
+        });
+        input = wrapper.find('input');
+    });
+
+    it('Set value', () => {
+        input.setValue('Amsterdam');
```
renancarvalho, line 20: If theres no business logic on setting a value on the input.. this test is not necessary.

src/__tests__/ui-input.test.js
```diff
@@ -0,0 +1,46 @@
...
+        expect(input.element.value).toBe('Amsterdam');
+    });
+
+    it('Is valid', async () => {
```
renancarvalho, line 24: I would rewrite to something like.. "Should add invalid class if the component state is invalid..."

src/__tests__/ui-input.test.js
larsdouweschuitema: - [ ] Remove

src/__tests__/ui-input.test.js
larsdouweschuitema: - [ ] Use more descriptive text in test descriptions

renancarvalho (dismissed):

renancarvalho (approved):

PR #424, opened 2021-05-26
feat(tooltip): make popper options versatile

larsdouweschuitema:
Make the tooltip (popper) options versatile instead of offering strict placement support.

Also deleted the key because I don't think it offers any value.

### DIFF
--- a/src/stories/ui-tooltip.stories.js
+++ b/src/stories/ui-tooltip.stories.js
@@ -16,10 +16,9 @@ export default {
                 options: ['hover', 'clickToOpen', 'clickToToggle', 'focus'],
             },
         },
-        side: {
+        options: {
             control: {
-                type: 'select',
-                options: ['top', 'right', 'bottom', 'left'],
+                type: 'object',
             },
         },
     },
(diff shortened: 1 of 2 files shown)


Zishrodrigues (approved):

src/stories/ui-tooltip.stories.js
```diff
@@ -41,8 +40,15 @@ const Template = (args, { argTypes }) => ({
 export const Tooltip = Template.bind({});
 Tooltip.args = {
     background: 'light',
-    trigger: 'hover',
-    side: 'bottom',
+    trigger: 'click',
+    options: {
```
dine, line 44: Maybe we can add the link to the vue-popperjs docs somewhere in this stories file? https://github.com/RobinCK/vue-popper#props

src/stories/ui-tooltip.stories.js
```diff
@@ -28,7 +27,7 @@ export default {
 const Template = (args, { argTypes }) => ({
     props: Object.keys(argTypes),
     components: { UiTooltip },
-    template: `<ui-tooltip :trigger="trigger" :side="side" :background="background">
+    template: `<ui-tooltip :trigger="trigger" :options="options" :background="background">
```
dine, line 30: This will be a breaking change

dine (changes requested):
can you add the "BREAKING CHANGE" message to the commit body?

dine (changes requested):
can you add the "BREAKING CHANGE" message to the commit body?

src/stories/ui-tooltip.stories.js
larsdouweschuitema: - [ ] make breaking change

src/stories/ui-tooltip.stories.js
larsdouweschuitema: - [ ] add link to doc

dine (approved):

PR #454, opened 2021-06-17
Add v-model support to input and select components

dine:


larsdouweschuitema (commented):
If we're going to add specific code related to making vuelidate work we might as well add an example using vuelidate.
Just have to make sure its a dev only dependency, so we don't add it to the library itself when installing.

bartderijk:
> If we're going to add specific code related to making vuelidate work we might as well add an example using vuelidate.
> Just have to make sure its a dev only dependency, so we don't add it to the library itself when installing.

I reckon the ability to properly use two-way data binding with v-model is not specific for use with Vuelidate. Altough the case was brought up by that situation, it's not limited to that scenario.

src/components/ui-input.vue
```diff
@@ -33,6 +35,13 @@ export default {
             type: Boolean,
             default: true,
         },
+        value: {
+            type: String,
+            default: '',
+        },
+    },
+    data: function () {
+        return { inputValue: this.value };
```
TobbenTM, line 44: Like it or not, a use case we also have is that we want input elements to react to other changes to the value. This won't work then, right?

IE:
```vue
<template>
    <ui-input v-model="exampleValue" />
</template>

<script>
export default {
  data () {
    return {
      exampleValue: null
    }
  },
  mounted () {
    this.exampleValue = 'this won't show up in the input'
  }
}
```

Am I wrong about this? Is this something we want this to support?

TobbenTM: Like, I thought having 

```
v-bind="$attrs"
v-on="$listeners"
```

Would be enough to have a working v-model?

PR #455, opened 2021-06-18
fix(input): add v-model support

dine:
https://vuejs.org/v2/guide/components.html#Using-v-model-on-Components

### DIFF
--- a/src/__tests__/ui-input.test.js
+++ b/src/__tests__/ui-input.test.js
@@ -38,4 +38,10 @@ describe('UiInput', () => {
         expect(wrapper.text()).toContain(SLOT_SUFIX);
         expect(input.classes()).toContain('pr-12');
     });
+
+    it('Should emit input event with value', () => {
+        const value = 'test-emit-event';
+        input.setValue(value);
+        expect(wrapper.emitted().input[0]).toEqual([value]);
+    });
 });
--- a/src/components/ui-input.vue
+++ b/src/components/ui-input.vue
@@ -12,7 +12,7 @@
                 'pr-12': hasSufix,
             }"
             v-bind="$attrs"
-            v-on="$listeners"
+            v-on="inputListeners"
         />
         <template v-if="hasSufix">
             <slot name="sufix"></slot>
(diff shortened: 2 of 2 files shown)


src/components/ui-input.vue
```diff
@@ -35,6 +35,14 @@ export default {
...
     },
     computed: {
+        inputListeners: function () {
+            var vm = this;
+            return Object.assign({}, this.$listeners, {
+                input: function (event) {
+                    vm.$emit('input', event.target.value);
+                },
+            });
+        },
```
bartderijk, line 45: Nice! Could be tweaked a little like so but that's just nitpicking
`inputListeners() {
            const vm = this;
            return Object.assign({}, this.$listeners, {
                input(event) {
                    vm.$emit('input', event.target.value);
                },
            });
        },`

TobbenTM: I'm a bit confused, what does this change actually achieve?
If you have a `<ui-input v-model="inputValue" />`, it'll do the exact same thing with or without this change?

TobbenTM: Maybe this is a very good use case for a unit test or two?

dine: @TobbenTM It's not the same, so I thought this change could be useful.
https://vuejs.org/v2/guide/components.html#Using-v-model-on-Components
Screenshot 2021-06-18 at 16 20 44

src/__tests__/ui-input.test.js
```diff
@@ -38,4 +39,11 @@ describe('UiInput', () => {
         expect(wrapper.text()).toContain(SLOT_SUFIX);
         expect(input.classes()).toContain('pr-12');
     });
+
+    it('Should emit input event with value', () => {
+        vm.$emit = jest.fn();
```
TobbenTM, line 44: Maybe personal opinion, but I'd rather avoid mocking stuff if we don't have to.
This test can be verified by waiting a tick (`await vm.$nextTick()`), then `expect(wrapper.emitted().input[0]).toEqual([value])`.
Thoughs?

dine: No preference, so I updated the PR

TobbenTM (approved):
👌

TobbenTM (approved):

PR #464, opened 2021-06-25
Update funda icons to v2

daniarsuprianto:

PR #467, opened 2021-06-28
feat(icons): remove ui-icon component, use icons from @funda/icons

dine:
BREAKING CHANGE: the `<ui-icon>` component is removed


daniarsuprianto:
Why is it separated from https://github.com/funda-frontend/ui/pull/464?

PR #486, opened 2021-07-08
feat(modal_close_button): add text option for close modal button

renancarvalho:
Add the possibility to pass any text to replace the default 'X' icon to close the modal.

### DIFF
--- a/src/components/ui-modal.vue
+++ b/src/components/ui-modal.vue
@@ -25,7 +25,7 @@
                     <button
                         v-if="showCloseIcon"
                         :class="[
-                            'focus:outline-none absolute w-11 h-11 top-1.5 right-1.5 p-2 text-blue-2 hover:text-blue-1',
+                            'focus:outline-none focus:text-blue-1 absolute w-11 h-11 top-1.5 right-1.5 p-2 text-blue-2 hover:text-blue-1',
                             closeButtonTextClass,
                         ]"
                         @click="close"
--- a/src/components/ui-modal.vue
+++ b/src/components/ui-modal.vue
@@ -101,7 +101,7 @@ export default {
     },
     computed: {
         closeButtonTextClass() {
-            return this.$props.closeButtonText ? 'w-auto' : '';
+            return this.$props.closeButtonText ? 'w-auto pr-3' : '';
         },
     },
     beforeMount() {
--- a/src/components/ui-modal.vue
+++ b/src/components/ui-modal.vue
@@ -24,10 +24,17 @@
                     </h2>
                     <button
                         v-if="showCloseIcon"
-                        class="focus:outline-none absolute w-11 h-11 top-1.5 right-1.5 p-2"
+                        :class="[
+                            'focus:outline-none absolute w-11 h-11 top-1.5 right-1.5 p-2 text-blue-2 hover:text-blue-1',
+                            closeButtonTextClass,
+                        ]"
                         @click="close"
                     >
-                        <icon-close class="text-blue-2" />
+                        <icon-close
+                            v-if="!closeButtonText"
+                            :title="closeButtonTitle"
+                        />
+                        {{ closeButtonText }}
                     </button>
                 </header>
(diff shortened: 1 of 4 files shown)


src/components/ui-modal.vue
```diff
@@ -24,10 +24,17 @@
                     </h2>
                     <button
                         v-if="showCloseIcon"
-                        class="focus:outline-none absolute w-11 h-11 top-1.5 right-1.5 p-2"
+                        :class="[
+                            'focus:outline-none absolute w-11 h-11 top-1.5 right-1.5 p-2 text-blue-2 hover:text-blue-1',
```
larsdouweschuitema, line 28: add focus too, so when you tap to the button it will show the same color as on hover.
removing the outline can remain so you dont have this weird default outline

larsdouweschuitema (changes requested):

src/stories/ui-modal.stories.js
```diff
@@ -14,6 +14,13 @@ export default {
                 options: [true, false],
             },
         },
+        closeButtonText: {
```
dine, line 17: Let's add the `closeButtonTitle` too

src/components/ui-modal.vue
renancarvalho: done.

src/stories/ui-modal.stories.js
```diff
@@ -84,6 +91,7 @@ const Template = (args, { argTypes }) => ({
             :customModalContainerClasses="customModalContainerClasses"
             :extraModalClasses="extraModalClasses"
             :showCloseIcon="showCloseIcon"
+            :closeButtonText="closeButtonText"
```
dine, line 94: Pass the `closeButtonTitle` also and  add both to the `TemplateModalWithContent` below too

src/components/ui-modal.vue
```diff
@@ -24,10 +24,17 @@
...
+                            closeButtonTextClass,
+                        ]"
                         @click="close"
                     >
-                        <icon-close class="text-blue-2" />
+                        <icon-close
+                            v-if="!closeButtonText"
+                            :title="closeButtonTitle"
+                        />
+                        {{ closeButtonText }}
```
dine, line 37: Let's check with @funda-frontend/ux if the padding remains the same when using text here. 
It looks a little bit too close to the right side for me:
https://funda-frontend.github.io/ui/refs/pull/486/merge/?path=/story/layout-modal--modal-default
Screenshot 2021-07-08 at 12 06 15

renancarvalho: Can make it pr-3 will look like this 
image

src/stories/ui-modal.stories.js
renancarvalho: done, next push will have it

src/stories/ui-modal.stories.js
renancarvalho: done, next push will have it

jpsc:
Hello @renancarvalho. While I understand that this PR solves the current limitation I believe that there is a better way to solve this. We currently have 3 properties that are for solving usage of the close button. Maybe we should think about [inversion of control](https://www.youtube.com/watch?v=AiJ8tRRH0f8&list=PLV5CVI1eNcJgNqzNwcs4UKrlJdhfDjshf) and giving the consuming user to define what is the close button and what does it do.  In this specific case, it would be to expose the 'close modal event'.

larsdouweschuitema (approved):

arjanvanrees:
In the examples in the storybook I notice that the textual 'Close' button is shown as default. I would like to request to change that to the X icon instead. The X icon used the most everywhere on the platform and should be the default in my opinion. Is that still possible?

renancarvalho:
@arjanvanrees, makes sense. remove the slots from the stories and the X will be there by default.

jpsc (dismissed):

dine (approved):

PR #490, opened 2021-07-16
chore(commitlint): add commitlint to validate commit messages

dine:


renancarvalho:
I tested and works well, we just have to double check how to handle sourcetree and these kinds of git Gui... 
It checked on @ConnorChristensen 's computer and it fails... we did a quick research and normally the problem is because the PATH variable on these GUi are not the same as the terminal... maybe it will have to be manual step to be done ... maybe we can add it on the readme.md

renancarvalho (commented):
Tried a commit with any random message and it did not accepted... good =)

dine:
New PR bumping also husky version https://github.com/funda-frontend/ui/pull/491

PR #491, opened 2021-07-16
chore(pkgs): bump husky and add commitlint to validate commit messages

dine:


ConnorChristensen (approved):
I gave this a test locally. As long as I run `npx husky install` while setting up the project, everything works well.

dine:
> I gave this a test locally. As long as I run `npx husky install` while setting up the project, everything works well.

@ConnorChristensen I just added a `postinstall` script to run `husky install`, just like we did for the icons:
https://github.com/funda-frontend/icons/blob/master/package.json#L57
Let me know if that works now.

ConnorChristensen (approved):
The post-install script works like a charm. Ship it!

PR #539, opened 2021-09-03
Automating dependabot

dine:
https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/automating-dependabot-with-github-actions


jpsc (approved):
Worth trying

bartderijk (approved):

PR #551, opened 2021-09-06
chore(workflows): build before deploy latest

dine:


ConnorChristensen (approved):
Looks reasonable to me

PR #557, opened 2021-09-10
fix(input)!: allow variable size in prefix and suffix

dine:
BREAKING CHANGE: Prefix and suffix should no longer have absolute
positioning.


ConnorChristensen:
Form what I have heard, there are UI changes coming that will likely impact this project and cause breaking changes. If this package is going to go through a breaking change, I would like to include all the breaking changes in one go. For now we can keep this PR open, but do not merge this until we have an understanding of what needs to go through breaking changes in this package in order to move us forward with the next version of Funda UI.

dine:
Closing this PR for now then

PR #561, opened 2021-09-17
feat: add ui-icon-button

bartderijk:
This component allows icons to be buttons on their own, all code credit goes to Connor.

### DIFF
--- a/src/components/ui-icon-button.vue
+++ b/src/components/ui-icon-button.vue
@@ -1,0 +1,39 @@
+<template>
+    <a v-if="showLink" :href="href" :class="cssClasses">
+        <slot />
+    </a>
+    <button v-else :class="cssClasses" :disabled="disabled" v-on="$listeners">
+        <slot />
+    </button>
+</template>
+
+<script>
+export default {
+    props: {
+        href: {
+            type: String,
+            default: null,
+        },
+        disabled: {
+            type: Boolean,
+            default: false,
+        },
+    },
+    computed: {
+        cssClasses() {
+            const enabledClasses =
+                'text-blue-2 hover:text-blue-1 cursor-pointer';
+            const disabledClasses = 'text-light-1 cursor-default';
+            return `leading-6 inline-flex whitespace-no-wrap border-solid rounded-sm align-middle text-center items-center justify-center focus:outline-none ${
+                this.disabled ? disabledClasses : enabledClasses
+            } `;
+        },
+        showLink() {
+            // if the button is disabled, we want to switch to showing a button
+            // so the user can not click the disabled icon and still follow the
+            // link.
+            return this.href && this.href.length > 0 && !this.disabled;
+        },
+    },
+};
+</script>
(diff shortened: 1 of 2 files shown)


src/components/ui-icon-button.vue
```diff
@@ -0,0 +1,39 @@
...
+            const disabledClasses = 'text-light-1 cursor-default';
+            return `leading-6 inline-flex whitespace-no-wrap border-solid rounded-sm align-middle text-center items-center justify-center focus:outline-none ${
+                this.disabled ? disabledClasses : enabledClasses
+            } `;
+        },
+        showLink() {
+            // if the button is disabled, we want to switch to showing a button
+            // so the user can not click the disabled icon and still follow the
+            // link.
+            return this.href && this.href.length > 0 && !this.disabled;
```
bartderijk, line 35: syntactic suger: `return this.href?.length > 0 && !this.disabled;`

src/components/ui-icon-button.vue
```diff
@@ -0,0 +1,39 @@
...
+            default: null,
+        },
+        disabled: {
+            type: Boolean,
+            default: false,
+        },
+    },
+    computed: {
+        cssClasses() {
+            const enabledClasses =
```
bartderijk, line 24: maybe we can clean this up a bit and story the values of `enabledClasses`, `disabledClasses ` and the long string value with all the classes in `data`, so the computed prop here only does the actual logic to determine what value to send

dine:
Why can't we use ui-button for icons? What is the main difference?

src/components/ui-icon-button.vue
ConnorChristensen: Good point. I will clean this up.

ConnorChristensen:
That is a good question. I actually forgot why I did it and almost scrapped the pull request. There are a few reasons I created this (at least for the project I was working on).

1. The button has a fixed height of `h-11` which adds vertical spacing around an icon and makes it less flexible in where it can be placed on the site (this was actually a deal breaker for what I was trying to implement in my project).
2. There are a lot of classes in the ui-button component that are not needed here and distract from the purpose of the icon button.
   * The icon does not need to be centered because the container should not be larger than the icon
   * The icon button does not need a rounded border because it does not have a border
   * etc.
3. There is an accessibility issue with the `focus:outline-none` for users that need to navigate the site via keyboard.
4. There is a bug on the ui-button component where a disabled link can still be clicked.
5. I usually do not do things just because others are doing it too, but the Material UI library is a pretty good reference when making design decisions and they have both a [Button](https://mui.com/api/button/) and an [IconButton](https://mui.com/api/icon-button/). They may have done this for some of the reasons mentioned above, or maybe it is just good developer experience to have a Button that assumes text and an IconButton.

While the bugs in ui-button should also be fixed, fixing those issues may not entirely eliminate the need for an icon button.

Upon more reflection though, I think it may be better based on the Vue style guide rule for [order of words in component names](https://vuejs.org/v2/style-guide/#Order-of-words-in-component-names-strongly-recommended) to name this component `<ButtonIcon>` and the component `ui-button-icon.vue`.

PR #562, opened 2021-09-17
feat(card-image): add the option for a small image

bartderijk:
Next to the full width images we're already supporting with the image card component, we're seeing another pattern in the UI where we have a smaller image in the top of the card. On small screens this should cover the entire width of the screen and on desktop it should behave like a normal 'small' image in the content.

To support this I've extended the card image component. By default it will always show the image that was passed full width like we already did. If the option (prop) `size` is set, it will show the smaller variant.

### DIFF
--- a/src/components/ui-card-image.vue
+++ b/src/components/ui-card-image.vue
@@ -1,13 +1,22 @@
 <template>
     <ui-card :variant="variant">
         <template #image>
-            <div class="overflow-hidden relative h-0 pt-40 md:rounded-t-sm">
+            <div
+                v-if="size === 'big'"
+                class="overflow-hidden relative h-0 pt-40 md:rounded-t-sm"
+            >
                 <img
                     class="absolute block w-full h-full top-0 left-0 object-cover"
                     v-bind="$attrs"
                     loading="lazy"
                 />
             </div>
+            <img
+                v-else
+                class="block px-0 md:px-4 pt-0 md:pt-4 w-full md:w-auto rounded-t-sm md:rounded-none"
+                v-bind="$attrs"
+                loading="lazy"
+            />
         </template>
         <slot />
     </ui-card>
@@ -25,6 +34,10 @@ export default {
         variant: {
             type: String,
             default: 'default',
+        },
+        size: {
+            type: String,
+            default: 'big',
         },
     },
 };
--- a/src/components/ui-card-image.vue
+++ b/src/components/ui-card-image.vue
@@ -1,8 +1,14 @@
 <template>
     <ui-card :variant="variant">
         <template #image>
+            <img
+                v-if="smallImage"
+                class="block px-0 md:px-4 pt-0 md:pt-4 w-full md:w-auto rounded-t-sm md:rounded-none"
+                v-bind="$attrs"
+                loading="lazy"
+            />
             <div
-                v-if="size === 'big'"
+                v-else
                 class="overflow-hidden relative h-0 pt-40 md:rounded-t-sm"
             >
                 <img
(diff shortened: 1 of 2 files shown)


src/components/ui-card-image.vue
```diff
@@ -26,6 +35,10 @@ export default {
             type: String,
             default: 'default',
         },
+        size: {
+            type: String,
```
bartderijk, line 39: I'll make this a Boolean

ConnorChristensen:
Looks good to me. The only improvement I see is that we can simplify the story a little. This should be all we need for the card image component, and anyone looking at the story can toggle the `smallImage` value using the storybook control.

```
const TemplateImage = (args, { argTypes }) => ({
    props: Object.keys(argTypes),
    components: { UiCardImage },
    template: ` <ui-card-image v-bind="$props">
                    Card content
                </ui-card-image>`,
});

export const CardImage = TemplateImage.bind({});
CardImage.args = {
    variant: 'default',
    smallImage: false,
    src: 'https://placekitten.com/300/200',
    alt: 'A kitty cat',
};
```

If you want to keep the CardImage and the CardImageSmall as two separate stories, we can add this on to the code block above. Our users do not have to touch the controls, but it is nice to give them the option.

```
export const CardImageSmall = TemplateImage.bind({});
CardImageSmall.args = {
    ...CardImage.args,
    smallImage: true,
};
```

src/components/ui-card-image.vue
```diff
@@ -1,7 +1,16 @@
 <template>
     <ui-card :variant="variant">
         <template #image>
-            <div class="overflow-hidden relative h-0 pt-40 md:rounded-t-sm">
+            <img
+                v-if="smallImage"
+                class="block px-0 md:px-4 pt-0 md:pt-4 w-full md:w-auto rounded-t-sm md:rounded-none"
+                v-bind="$attrs"
+                loading="lazy"
```
ConnorChristensen, line 8: Any reason why this is being lazy loaded and the full size image is not?

ConnorChristensen (approved):
There are a few comments above if you have the time, but I did not find anything concerning.

src/components/ui-card-image.vue
bartderijk: None, they both have it :) (expand the code block to see all the code)

bartderijk:
> Looks good to me. The only improvement I see is that we can simplify the story a little. This should be all we need for the card image component, and anyone looking at the story can toggle the `smallImage` value using the storybook control.
> 
> ```
> const TemplateImage = (args, { argTypes }) => ({
>     props: Object.keys(argTypes),
>     components: { UiCardImage },
>     template: ` <ui-card-image v-bind="$props">
>                     Card content
>                 </ui-card-image>`,
> });
> 
> export const CardImage = TemplateImage.bind({});
> CardImage.args = {
>     variant: 'default',
>     smallImage: false,
>     src: 'https://placekitten.com/300/200',
>     alt: 'A kitty cat',
> };
> ```
> 
> If you want to keep the CardImage and the CardImageSmall as two separate stories, we can add this on to the code block above. Our users do not have to touch the controls, but it is nice to give them the option.
> 
> ```
> export const CardImageSmall = TemplateImage.bind({});
> CardImageSmall.args = {
>     ...CardImage.args,
>     smallImage: true,
> };
> ```

Done, thanks!

src/components/ui-card-image.vue
ConnorChristensen: Ah, my bad!

ConnorChristensen (approved):

PR #572, opened 2021-09-28
fix(notification): fix icon alignment for multiline message

dine:


bartderijk (dismissed):
What Connor said, +approved :)

ConnorChristensen:
Is this new abstraction of the ui-icon-type an internal only change, or is this also a feature that should be used by other projects?

dine:
image

I could add some stories for the ui-icon-type too

ConnorChristensen:
I am ok with it either way, but if we do, I think it should be a feature rather than a fix and I would recommend removing the required margin right around the icon to make incorporating it into different designs more flexible.

bartderijk (approved):

PR #636, opened 2021-11-04
Feature/acc 132 switch toggle

ivanasetiawan:
In order to make it clear for user how to turn on/off various email settings in the email settings page we want to add a toggle.


src/components/ui-switch-toggle.vue
```diff
@@ -0,0 +1,37 @@
...
+        </label>
+    </div>
+</template>
+<script>
+export default {
+    props: {
+        id: {
+            type: String,
+            required: true,
+            default: '',
```
bartderijk, line 18: Since id is required we can drop the default

package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@funda/ui",
-    "version": "8.8.0",
+    "version": "8.9.0",
```
daniarsuprianto, line 3: the versioning is done automatically based on commit message, so, dont need to change it manually here :)

- [ ] please revert

src/stories/ui-switch-toggle.stories.js
```diff
@@ -0,0 +1,21 @@
...
+export default {
+    title: 'Forms/Checkbox/Switch toggle',
+    component: UiSwitchToggle,
+};
+
+const Template = (args, { argTypes }) => ({
+    props: Object.keys(argTypes),
+    components: { UiSwitchToggle },
+    template:
+        '<ui-switch-toggle :checked="true" id="switch-toggle-1" @change="onChange" />',
```
daniarsuprianto, line 12: i assumed `:checked="true"` should be `:checked="checked"` ?

- [ ] update

daniarsuprianto (changes requested):

src/stories/ui-switch-toggle.stories.js
```diff
@@ -0,0 +1,21 @@
+import UiSwitchToggle from '@/components/ui-switch-toggle.vue';
+
+export default {
+    title: 'Forms/Checkbox/Switch toggle',
```
daniarsuprianto, line 4: Even though this is a checkbox basically, i'd prefer to have this outside "Checkbox" group because it just feels different to me hahaha.... but i will leave it completely up to you

daniarsuprianto:
- [ ] Please check final design with rick

daniarsuprianto (changes requested):
- [ ] Please check final design with rick

ConnorChristensen (changes requested):

src/stories/ui-switch-toggle.stories.js
ConnorChristensen: If you name this `Switch Toggle` (the same name as the export const), then the storybook sidebar/navigator will look like this
Screen Shot 2021-11-05 at 10 20 41
 instead of this
Screen Shot 2021-11-05 at 10 19 46

daniarsuprianto (approved):

src/stories/ui-switch-toggle.stories.js
ivanasetiawan: I will make Switch Toggle independent.. I think it's more clear like that.

src/components/ui-switch-toggle.vue
```diff
@@ -0,0 +1,40 @@
...
+            required: true,
+        },
+        checked: {
+            type: Boolean,
+            default: false,
+        },
+    },
+    data() {
+        return {
+            isChecked: this.checked,
```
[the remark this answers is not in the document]
ivanasetiawan: Good catch!

src/stories/ui-switch-toggle.stories.js
```diff
@@ -0,0 +1,21 @@
...
+    component: UiSwitchToggle,
+};
+
+const Template = (args, { argTypes }) => ({
+    props: Object.keys(argTypes),
+    components: { UiSwitchToggle },
+    template:
+        '<ui-switch-toggle :checked="checked" :id="id" :name="name" @change="onChange""></ui-switch-toggle>',
```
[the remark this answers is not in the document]
ivanasetiawan: Personal preference ;)

jpsc (changes requested):
Hi @ivanasetiawan ,

I think you are missing a few things on the component to make it complete.
- There is no ::active styles (you don't know when the switch is selected)
- There is no keyboard action to activate - According to [this document](https://www.w3.org/TR/wai-aria-1.1/#switch) it needs to behave like a checkbox. 
- For the naming you probably need to choose one of the other. switch-toggle is redundant.
- I'm not sure if possible, but the `v-model` should be the value of checked why do we need both?
- Using the `mounted` lifecycle to set the state means that the initial state when coming from the ssr can be out of sync, anything to avoid this?

jpsc:
The controls of the story for the checked state don't seem to be working also.

src/stories/ui-switch-toggle.stories.js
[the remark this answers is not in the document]
ConnorChristensen: I think this actually makes an impact (even if it is small) to how stories are written. While we should trust our engineers to be careful, we can also remove the possibility of someone adding an arg and not adding it to the template, or removing a prop from the template and not adjusting the corresponding args. One can certainly overdo the Don't Repeat Yourself principle, I don't think this is one of those cases.

The storybook folks use this in their docs: https://storybook.js.org/docs/vue/get-started/whats-a-story

ivanasetiawan:
Hi @jpsc 

- There is no ::active styles (you don't know when the switch is selected)
> How do you mean? It's still WIP and I am applying disabled, focused and added checkbox icon based on Rick's design here https://www.figma.com/file/zEYQtitTwXrNgDuMQQPqp9/Toggle-switch?node-id=0%3A1 but there's already a "checked" style

- There is no keyboard action to activate - According to this document it needs to behave like a checkbox.
> On my local, when I hit `space` on my keyboard, it toggles on/off. I will take closely into it and will poke you if I still don't get what you mean

- For the naming you probably need to choose one of the other. switch-toggle is redundant.
> I will rename to Switch, I think it's more clear

- I'm not sure if possible, but the v-model should be the value of checked why do we need both?
> You're right, I actually removed it but haven't pushed 

- Using the mounted lifecycle to set the state means that the initial state when coming from the ssr can be out of sync, anything to avoid this?
> This question also applies to our `ui-accordion`. Have we encountered any issues with that? I am all ears for suggestions on how to approach this based on real case.

jpsc:
> On my local, when I hit `space` on my keyboard, it toggles on/off. I will take closely into it and will poke you if I still don't get what you mean

That's actually true. It didn't seem to work because of the previous point. There are no ::focus styles so you don't know when you are actually on the element.

src/stories/ui-switch-toggle.stories.js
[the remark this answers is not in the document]
ivanasetiawan: I will take a look at this closely, but `<ui-switch-toggle v-model="$props"></ui-switch-toggle>` is not working.

src/components/ui-switch.vue
```diff
@@ -0,0 +1,78 @@
+<template>
```
ivanasetiawan, line 1: Make sure the height is at least 44px (per Rick's request)

bartderijk (dismissed):

src/components/ui-switch.vue
```diff
@@ -0,0 +1,78 @@
...
+        }
+    },
+    data() {
+        return {
+            isChecked: false,
+        };
+    },
+    computed: {
+        cssBgClasses() {
+            return this.isChecked ? 'bg-blue-3 hover:bg-blue-1' : 'bg-light-1 hover:bg-dark-3';
```
ConnorChristensen, line 57: Hover state can not exist only here. Hover state still needs to be valid when hovering over the circle in the switch.

ConnorChristensen (changes requested):
Please consult the switch design again and make sure all states match the design.

src/components/ui-switch.vue
ivanasetiawan: Yes can with group-hover. Updated.

src/components/ui-switch.vue
```diff
@@ -0,0 +1,78 @@
...
+    <div class="flex items-center">
+        <label :for="id" class="h-11 flex items-center cursor-pointer focus:outline-none focus:ring-2 focus:ring-blue-3 focus:ring-offset-2 rounded-full">
+            <div class="relative">
+                <input
+                    type="checkbox"
+                    class="sr-only"
+                    :id="id"
+                    role="switch"
+                    :name="name"
+                    :disabled="disabled"
```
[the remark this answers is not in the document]
ivanasetiawan: Thank you for noticing this, I forgot to add the disabled style! Fixed :)

ConnorChristensen:
* Focus state is not working on toggle
* It would be nice if there was a transition to and from the hover state
* Bonus points if you get the check icon to fade in and out when toggling a icon switch (definitely not a requirement if you are in a rush)

After toggle state is fixed, I will give this an approval.

ivanasetiawan:
@ConnorChristensen made all the changes: focus + transition + delay check icon.

package.json
```diff
@@ -24,7 +24,8 @@
         "generateExports": "node ./src/generateExports.js",
         "test": "jest",
         "test-watch": "jest --watch",
-        "prepare": "husky install"
+        "prepare": "husky install",
+        "lint": "eslint --fix --ext \".js,.vue\" --ignore-path .gitignore ."
```
jpsc, line 28: Not sure why this is part of this PR but it's not a blocker for me.

ivanasetiawan: Yea I was running lint and thought I could just add it here. I don't think it hurts to leave it here.

ConnorChristensen:
Looks good! Last thing is all these commits need to be squished into a single commit that adds this new switch component. All of your commits contain the `feat` tag. On release this will add 17 feature bullet points to the changelog that contain mostly fixes, which will be very confusing for future readers.

P.S. If you also want to rebase the branch on master while you are squishing commits, that will be appreciated, but not required.

jpsc (dismissed):

ConnorChristensen (approved):
I see the mouse is still set to a pointer when hovering over a disabled switch, but this is small enough that it is not worth blocking this pull request anymore. If you feel like fixing it and adding a single `fix` commit, that would be appreciated, but we can do it later if need be.

ivanasetiawan:
Let's do it right! I made the changes with the hover state on disabled and squash the commit.

If you don't like my approach with sprinkling `cursor-pointer` on multiple elements, you can fix it (create a PR for it). 
(I have tested with only putting `cursor-pointer` on the `label` and I didn't like behaviour that's why I took this approach.)

daniarsuprianto (approved):

ConnorChristensen:
Looks good to me!

PR #637, opened 2021-11-05
Set up easy linting

ConnorChristensen:
While most people probably have ESlint integration into VScode or other editor, this is an easy way to check the whole project at once, and makes it more accessible for those folks that prefer the terminal to keep their work in-line with the linter.


ivanasetiawan (approved):

src/stories/ui-label.stories.js
```diff
@@ -5,7 +5,7 @@ export default {
     component: UiLabel,
 };
 
-const Template = args => ({
```
bartderijk, line 8: @ConnorChristensen  @ivanasetiawan I'm pretty sure this change causes our code examples in storybook to loose their formatting. While I really appreciate quick follow-ups on PR's, please hold off on merging until others had a chance to look at the changes as well. 10 minutes from opening the PR until merging is a little too eager :)

ivanasetiawan: @bartderijk You're right! I was way too eager.. My apology. Shall revert and re-create this branch?

bartderijk: No worries, I found that almost all code examples are broken. I'll create a PR to fix them all in one go.

ivanasetiawan: Thank you @bartderijk - I will keep this in mind next time.

jpsc:
Not really sure if this was required. 

Sure, it's just an added script and it's minimal in impact but we are doing the linting and fixing on commit automatically, so I'm not sure why would you need it.

src/stories/ui-label.stories.js
jpsc: > 10 minutes from opening the PR until merging is a little too eager :)

Yep 😇

ConnorChristensen: I think that my linting change does not break the formatting, but I would be happy to help investigate the cause. I also agree that there is no urgency to get a change like this merged in.

ConnorChristensen:
I was reviewing #636 and some of the committed code did not follow the linting rules, so I figured it could be a help to the team to be able to run the linter manually without having to use `npx`. It looks like the husky configuration on commit does have linting for commit messages, but no rules for linting the source code:

```sh
#!/bin/sh
. "$(dirname "$0")/_/husky.sh"

npx --no-install commitlint --edit "$1"
```

Maybe this was something the repo had but it has been turned off or broken?

jpsc:
Most likely. I've cloned the repo and encountered the same behavior. 
I would prefer to fix this than to have that script.

src/stories/ui-label.stories.js
bartderijk: Like discussed, indeed it was already broken. Thanks for checking, I'll look into a fix.

PR #645, opened 2021-11-18
feat(ui-button): allow inputs of all types

bartderijk:
We had a case where the `type` was passed along with several other attributes. In this case the `type` was not connected to the input, effectively not picking up the input `type`. This is happening because we are binding all passed arguments to the input, given that they are not declared as a prop.

Adding to that, the docs mention the support of all types of inputs. The component however does not support this in the above scenario.

For that reason I have remove the `type` prop all together, allowing all input types.


src/components/ui-input.vue
```diff
@@ -24,11 +24,6 @@
 export default {
     inheritAttrs: false,
     props: {
-        type: {
-            type: String,
-            default: 'text',
-            validator: (value) => ['text', 'number'].includes(value),
-        },
```
bartderijk, line 31: We should still provide a default `type`, I'll add that.

bartderijk: Nevermind, an input without the `type="text"` attribute defaults to `type="text"` with shear HTML magic on its own already.

ConnorChristensen:
I would consider this commit a `fix` rather than a `feat`

ivanasetiawan (dismissed):
Thank you @bartderijk for the quick action 🙏

larsdouweschuitema (approved):

messenjer:
Hello @bartderijk,

Why the solution is not to add the type in the markup?

```diff
        <input
            class="w-full rounded-sm border bg-white shadow-inner py-2 px-4 outline-none text-base font-normal h-11 disabled:bg-light-3"
+          :type="type"
            :class="{
                'border-light-1 focus:border-dark-3': isValid,
                'border-red-1 focus:border-red-1': !isValid,
                'pl-12': hasPrefix,
                'pr-12': hasSuffix,
            }"
            v-bind="$attrs"
            v-on="inputListeners"
        />
```

I think that fixes the problem.

dine:
Hey @messenjer, we are already binding all the attributes with `v-bind="$attrs"`, so why would we need to explicitly add the type in the markup?

messenjer:
Hi @dine, 

Yesterday, I was working on a vue3 version of the component, and I ran into this problem and solved with `:type="type"`. I was just thinking about reporting it to you this morning, and then I saw this pull request on your repository.

You can test it with type="number" or type="password" for example.

Without :type="type", the field are not working.

# with type="number" or type="password"

Screenshot 2021-11-19 at 12 26 23

# number with :type="type"

Screenshot 2021-11-19 at 12 23 07

# password with :type="type"

Screenshot 2021-11-19 at 12 22 52

dine:
Hi @messenjer, we are not removing the `:type="type"` from the stories file, just from the component:
https://github.com/funda-frontend/ui/blob/master/src/stories/ui-input.stories.js#L48

You can see it works on this branch:
https://funda-frontend.github.io/ui/refs/pull/645/merge/?path=/story/forms-input--input&args=type:password

dine (dismissed):

bartderijk:
> I would consider this commit a `fix` rather than a `feat`

Fixed!

bartderijk:
@messenjer Thanks for the suggestion,[ you mean like this? ](https://github.com/funda-frontend/ui/pull/645/commits/0da62b69596cfa7fd698962504724df3c6d6319c)

dine (approved):

messenjer:
> @messenjer Thanks for the suggestion,[ you mean like this? ](https://github.com/funda-frontend/ui/pull/645/commits/0da62b69596cfa7fd698962504724df3c6d6319c)

Yes, like that.

PR #651, opened 2021-11-23
fix(tooltip): change reference element

larsdouweschuitema:
BREAKING CHANGE: element changed from button to div

### Why this change?
Duplicate button when using your own component inside the slot.
Popper requires a slot named reference.
Not only will it render a duplicate button in the DOM, it also doesnt allow you to trigger the tooltip on button click due to duplicate elements.

https://www.npmjs.com/package/vue-popperjs


src/components/ui-tooltip.vue
```diff
@@ -9,9 +9,9 @@
                 <slot name="content" />
             </div>
 
-            <button slot="reference" class="focus:outline-none">
+            <div slot="reference">
```
dine, line 12: Do you think we should add a `class="cursor-pointer"` here now that the button is gone?

larsdouweschuitema:
I also noticed hover is not working, but thats regardless of using button or div. We should check this out in another issue.

dine (approved):

dine:
> I also noticed hover is not working, but thats regardless of using button or div. We should check this out in another issue.

I notice it doesn't work on the Canvas tab, but it does on Docs https://funda-frontend.github.io/ui/latest/?path=/docs/visual-tooltip--tooltip

src/components/ui-tooltip.vue
larsdouweschuitema: @dine If you for example use your own component e.g. UiButton it will have the cursor pointer on there.

larsdouweschuitema:
> > I also noticed hover is not working, but thats regardless of using button or div. We should check this out in another issue.
> 
> I notice it doesn't work on the Canvas tab, but it does on Docs https://funda-frontend.github.io/ui/latest/?path=/docs/visual-tooltip--tooltip

Interesting. I see the hover working also on the doc. Maybe its just a storybook issue?

ConnorChristensen:
I think this has the unintended side effect of no longer having a pointer on the reference, and the text can be highlighted.

PR #664, opened 2021-11-29
UI modal: add variants and restrict extra classes

daniarsuprianto:


src/stories/ui-modal.stories.js
```diff
@@ -14,102 +14,21 @@ export default {
...
-            },
-        },
-        noScroll: {
-            defaultValue: false,
-            description:
-                'Boolean indicating if the modal body content should scroll',
-            control: {
-                options: [true, false],
+                type: 'select',
+                options: ['default', 'halfpage', 'fullscreen'],
```
ConnorChristensen, line 20: I see a console warning that `control.options` is [deprecated and will be removed](https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#deprecated-controloptions) in storybook v7.

src/stories/ui-modal.stories.js
```diff
@@ -14,102 +14,21 @@ export default {
...
-    },
-};
-
-const TemplateModalWithContent = (args, { argTypes }) => ({
     props: Object.keys(argTypes),
     components: { UiButton, UiModal, UiNotification },
     data() {
         return {
-            isModalVisible: false,
+            isModalVisible: true,
```
ConnorChristensen, line 31: We can define this on the args and then let people control it from storybook. Then there is no need to have this data section.

src/stories/ui-modal.stories.js
```diff
@@ -126,85 +45,62 @@ const TemplateModalWithContent = (args, { argTypes }) => ({
         <ui-modal
```
ConnorChristensen, line 45: If you like, we can simplify this to `<ui-modal v-bind="$props" v-show="isModalVisible" @close="closeModal">`

src/stories/ui-modal.stories.js
```diff
@@ -126,85 +45,62 @@ const TemplateModalWithContent = (args, { argTypes }) => ({
...
-                </div>
+                <p class="px-12 pb-12">
+                    Lorem ipsum dolor sit amet, consectetur adipiscing elit. Quisque cursus dui ullamcorper, faucibus felis vel, gravida purus. Mauris vel sagittis mi, id vehicula ex. Proin in mi ut metus malesuada gravida. Quisque lobortis dolor justo, sed bibendum sem auctor a. Donec nec sapien non felis finibus suscipit et nec sem. Aliquam vestibulum auctor dui, eu tincidunt nisl ullamcorper sagittis. Morbi in ex velit.
+                </p>
             </template>
-            <template v-slot:footer>
-                <ui-button tone="primary" @click="closeModal" class="mr-4">Button</ui-button>
-                <ui-button tone="tertiary" @click="closeModal">Button</ui-button>
+            <template v-if="footer" v-slot:footer>
```
ConnorChristensen, line 61: If we change `footer` to `showFooter`, then we can include that as an arg in `Default.args` and let the storybook user control it. It is not showing up as a control currently because it is overloading the footer slot. Then we can have a separate story for modal with buttons, but it is no longer required to allow a storybook user to explore all the variants.

src/stories/ui-modal.stories.js
```diff
@@ -126,85 +45,62 @@ const TemplateModalWithContent = (args, { argTypes }) => ({
...
-            
-        </ui-modal>
-    </div>`,
-});
-
-export const ModalWithCustomCloseButton = TemplateModalWithCustomCloseButton.bind(
-    {}
-);
-ModalWithCustomCloseButton.parameters = {
+export const ModalWithNotification = Template.bind({});
```
ConnorChristensen, line 97: If we apply the same trick as the `showFooter`, we can remove the need for this story.

src/components/ui-modal.vue
```diff
@@ -67,30 +70,41 @@ export default {
...
+            return this.variant === 'halfpage';
         },
-        customModalClasses: {
-            type: String,
-            default:
-                'w-full m-auto bg-white md:rounded shadow-xl transition duration-300 ease-in-out relative overflow-hidden flex flex-col',
+        containerVariantClasses() {
+            if (this.fullscreen) return 'h-full w-full';
+
+            if (this.halfpage)
```
ConnorChristensen, line 95: I recommend against assuming that JavaScript will properly apply the conditional scope to a value on the next line. JavaScript automatic semicolon insertion could mess this up.

daniarsuprianto: do you mean to put it inside curly brackets?

src/components/ui-modal.vue
```diff
@@ -1,57 +1,60 @@
...
-            </div>
+    <transition
+        enter-class="opacity-0"
+        leave-active-class="opacity-0"
+        @enter="animateInContent"
+        @enter-cancelled="animateOutContent"
+        @leave="animateOutContent"
+    >
+        <div
+            class="fixed inset-0 flex items-center justify-center z-50 transition duration-500 ease-in-out bg-black absolute inset-0 bg-opacity-60"
```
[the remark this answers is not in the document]
daniarsuprianto: sharp

src/components/ui-modal.vue
```diff
@@ -1,57 +1,60 @@
...
+            :class="fullscreen ? 'md:p-9' : ''"
+            @click.self="close"
+        >
             <div
```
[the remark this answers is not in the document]
daniarsuprianto: not really. the reason for this `div` (i'll call parentDiv) is the main container of the modal, which holds the backdrop and the content. the next `div` (i'll call it childDiv) is there because for the default modal, the max-width of the content is based on grid, instead of using static max-width value. we have 12 grids in the container of 984px, which is defined by the `childDiv`, and the content of the modal if 8 grids from that. if I remove the `childDiv`, the 12 grids will be defined by the `parentDiv` which doesn't have max-width of 984px, which we don't want, because the 8 grids content will be wrong.
However, this won't be the case for fullscreen modal, because fullscreen modal is not using 8 grids as max-width. Therefore you see on `childDiv`, instead of defining static class, im using `containerVariantClasses`, which is computed value based on the modal variant.
Do you know what I mean and does it make sense?

src/components/ui-modal.vue
ConnorChristensen: Yep. The newline without the curly braces *probably* will not be a problem, but it could be, and I try to avoid situations like that.

Here is more information, including some common bugs: https://en.wikibooks.org/wiki/JavaScript/Automatic_semicolon_insertion

src/components/ui-modal.vue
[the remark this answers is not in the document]
ConnorChristensen: Forgive my pseudocode here, but is why can the size of the modal not be defined by the following?

### default
* md:648px max width 
* md:65.5% width
* sm:100% width

### Half page
* md:65.5% width
* sm:100% width

### full screen
* md:95% width
* sm:100% width

In the default modal case, the childDiv is already set to a max width of 984px, so why not just apply a max width to the modal instead of making the modal a fraction of a max width?

If this is just how it is, I am ok with it, but it seems overly complex to me.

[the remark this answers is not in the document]
daniarsuprianto: it can't, according to UX, modal width is not by pixels, but by grids. 8 grid out of 12. for example, on 1024px screen, the modal will be 648px, but if 768, it will be 500 something, depends on the grid.. and also 8 grids out of 12 is not exactly 3/4 or 75% of the screen either. so it's not defined by the pixels/percentage.

Zishrodrigues (approved):
I like how you did the variants; clear & clean.

PR #686, opened 2021-12-06
fix(modal): fix fullscreen variant height

daniarsuprianto:


src/__tests__/ui-modal.test.js
```diff
@@ -5,7 +5,7 @@ const BUTTON_SELECTOR = 'button';
 const ATTR_MODAL_CONTAINER = '[data-test-modal-container]';
 const ATTR_MODAL = '[data-test-modal]';
 const DEFAULT_MODAL_CLASSES =
-    'bg-white shadow-xl relative flex flex-col md:block md:max-h-full';
+    'bg-white shadow-xl relative flex flex-col md:max-h-full';
```
ConnorChristensen, line 8: I think the test needing to be updated to fix a visual bug is an indication that the test is not very flexible and the test was not targeting the right thing in the first place. It is hard to do a visual test through unit testing classes on a component.

This test is effectively a manual implementation of a snapshot, which checks if the output has changed, not if it was correct. I suggest we either remove the visual unit tests, or we implement snapshot testing, and would like to know if others agree with me.

ConnorChristensen (changes requested):
There appear to still be some visual bugs.

* A modal with long body content can overflow the screen height on short browser windows.
* A modal with long body content with `{ mobileFullScreen: true, variant: 'fullscreen' }` has a weird issue where offscreen content is hidden behind the background when you scroll down.
* It may be worth giving this storybook story to a quality assurance person, as there are a lot of possible variations and I do not think I covered all of them in my review of this pull request.

While we are at it, lets make a few adjustments to the modal story:

* If you are only going to have one story, rename `Default` to `Modal` and it will prevent the story from being a folder with a single story (less clicking to navigate between stories).
* Now that there is no more state in the story that gives a visual indication that the modal actually closes, we should connect actions so we can see that these buttons on the modal actually fire the close event.
  * [This github issue](https://github.com/storybookjs/storybook/discussions/11372) shows how to connect actions.
  * [This storybook documentation page](https://storybook.js.org/docs/vue/essentials/actions) looks like it is showing how to connect actions in the new version 6.4. This might work for our current version but it did not work for me when I gave it a quick try.

src/components/ui-modal.vue
```diff
@@ -9,27 +9,31 @@
...
-                    {
-                        'h-screen md:h-auto overflow-auto md:overflow-visible': mobileFullscreen,
-                    },
-                ]"
+                class="w-full"
+                :class="{
+                    'h-full md:h-full': isFullscreen,
+                    'h-full md:grid gap-x-6 grid-cols-12': isHalfpage,
+                    'h-auto max-w-container md:grid gap-x-6 grid-cols-12':
+                        !isFullscreen && !isHalfpage,
```
ConnorChristensen, line 22: This conditional should be a computed value.

src/components/ui-modal.vue
```diff
@@ -9,27 +9,31 @@
...
-                    class="bg-white shadow-xl relative flex flex-col md:block md:max-h-full"
+                    class="bg-white shadow-xl relative flex flex-col h-full overflow-auto md:max-h-full"
                     :class="[
-                        variantClasses,
                         animationClasses,
                         {
+                            'col-start-3 col-span-8': isDefault,
+                            'col-start-4 col-span-6': isSmall,
+                            'col-span-8': isHalfpage,
+                            'md:rounded': !isHalfpage,
```
ConnorChristensen, line 36: `isHalfpage ? 'col-span-8' :  'md:rounded'`

src/components/ui-modal.vue
```diff
@@ -115,37 +119,17 @@ export default {
         };
     },
     computed: {
-        fullscreen() {
+        isFullscreen() {
```
ConnorChristensen, line 122: I like the rename to make it easy to see that it returns a boolean!

ConnorChristensen (dismissed):
I think the body of the modal should always scroll if the screen gets too small. Still seeing some cases where the modal expands past the width of the screen.

src/stories/ui-modal.stories.js
```diff
@@ -35,26 +35,38 @@ export default {
...
-                </p>
-                <p class="px-4 pb-4 md:px-6" v-if="longContent">
-                    Lorem ipsum dolor sit amet, consectetur adipiscing elit. Quisque cursus dui ullamcorper, faucibus felis vel, gravida purus. Mauris vel sagittis mi, id vehicula ex. Proin in mi ut metus malesuada gravida. Quisque lobortis dolor justo, sed bibendum sem auctor a. Donec nec sapien non felis finibus suscipit et nec sem. Aliquam vestibulum auctor dui, eu tincidunt nisl ullamcorper sagittis. Morbi in ex velit.
-                </p>
-                <p class="px-4 pb-6 md:px-6">
-                    Lorem ipsum dolor sit amet, consectetur adipiscing elit. Quisque cursus dui ullamcorper, faucibus felis vel, gravida purus. Mauris vel sagittis mi, id vehicula ex. Proin in mi ut metus malesuada gravida. Quisque lobortis dolor justo, sed bibendum sem auctor a. Donec nec sapien non felis finibus suscipit et nec sem. Aliquam vestibulum auctor dui, eu tincidunt nisl ullamcorper sagittis. Morbi in ex velit.
-                </p>
+                <div :class="{'max-h-96 overflow-auto': longContent && (isDefault || variant === 'small')}">
```
bartderijk, line 53: Another great candidate for a computed prop

src/stories/ui-modal.stories.js
```diff
@@ -35,26 +35,38 @@ export default {
...
+                        Lorem ipsum dolor sit amet, consectetur adipiscing elit. Quisque cursus dui ullamcorper, faucibus felis vel, gravida purus. Mauris vel sagittis mi, id vehicula ex. Proin in mi ut metus malesuada gravida. Quisque lobortis dolor justo, sed bibendum sem auctor a. Donec nec sapien non felis finibus suscipit et nec sem. Aliquam vestibulum auctor dui, eu tincidunt nisl ullamcorper sagittis. Morbi in ex velit.
+                    </p>
+                    <p class="px-4 pb-6 md:px-6">
+                        Lorem ipsum dolor sit amet, consectetur adipiscing elit. Quisque cursus dui ullamcorper, faucibus felis vel, gravida purus. Mauris vel sagittis mi, id vehicula ex. Proin in mi ut metus malesuada gravida. Quisque lobortis dolor justo, sed bibendum sem auctor a. Donec nec sapien non felis finibus suscipit et nec sem. Aliquam vestibulum auctor dui, eu tincidunt nisl ullamcorper sagittis. Morbi in ex velit.
+                    </p>
+                </div>
             </template>
             <template v-if="showFooter" v-slot:footer>
                 <ui-button tone="primary" class="mr-4">Save</ui-button>
```
bartderijk, line 72: Not a change, I know, but `primary` is the default tone so we can omit it here.

bartderijk:
I missed the context that led up to this PR as I was off for a while. Adding a description, even a tiny one, would be super helpful.

src/__tests__/ui-modal.test.js
daniarsuprianto: i did snapshot tests yeah.. i agree with you that checking the class is not the best way

src/components/ui-modal.vue
daniarsuprianto: i prefer doing it in different line so that it matches with the rest of the conditional class, just to make it easier more neat imo. so i will leave it how it is if you dont mind :)

daniarsuprianto:
> A modal with long body content can overflow the screen height on short browser windows.
A modal with long body content with { mobileFullScreen: true, variant: 'fullscreen' } has a weird issue where offscreen content is hidden behind the background when you scroll down.

@bartderijk mostly because of the bugs Connor's found because of the previous breaking change when we decided to remove "customModalClasses" property and make variants instead

src/components/ui-modal.vue
```diff
@@ -9,27 +9,31 @@
         <div
             class="fixed inset-0 flex justify-center items-center z-50 transition duration-500 ease-in-out bg-black bg-opacity-60"
             :class="{
-                'md:p-9': fullscreen,
+                'md:p-9': isFullscreen,
```
larsdouweschuitema, line 12: What did you think of the branch I shared earlier?

daniarsuprianto: i like the idea of making everything inside modal as a slot.. however im in doubt if we should have different component like `modal-header` or `modal-footer` or `modal-content`.. i mean, not sure if that's necessary. but im 100% make a slot for the whole content

src/components/ui-modal.vue
```diff
@@ -115,37 +119,17 @@ export default {
         };
     },
     computed: {
-        fullscreen() {
+        isFullscreen() {
             return this.variant === 'fullscreen';
         },
-        halfpage() {
+        isHalfpage() {
```
larsdouweschuitema, line 125: Small detail but we talk about "screen" in 1 property and "page" in the other.

daniarsuprianto: if "halfscreen" is more common word, im open to change it. wdyt? is it halfpage or halfscreen?

ivanasetiawan: This depends on what we wanna show. `Screen` is different than `Page`. 

If we use `halfpage`, I would assume it will be half of the page - so if it's a long page, then this modal will be in the middle, which could be out of the screen.. I'd vote for `halfscreen` IF the purpose is to show in the middle of the screen :)

ivanasetiawan (approved):
Regarding my comment for `Page` vs. `Screen`, please consider but feel free to use as you see fit :)

src/components/ui-modal.vue
daniarsuprianto: Discussed with Ivana and we agree to rename this to halfscreen. However, it will be breaking change renaming this, i'd suggest to do it on the next branch about ui-modal to make the content of the modal as a slot. that way we only have 1 version with breaking change instead of 2

daniarsuprianto:
> I think the body of the modal should always scroll if the screen gets too small. Still seeing some cases where the modal expands past the width of the screen.

it should be handled by adding max-height to the content imo. thing is if user choose not to use the "mobileFullscreen" option, the modal should stay in the middle. we can check "if content is too long, then modal align to top, not in the middle and make it fullscreen instead", but i wouldnt want to add javascript to check the content height to determine the alignment. or they can just use option "mobileFullscreen"

PR #689, opened 2021-12-08
Improve the focus-visible state of the button

bartderijk:
This PR introduces improved a11y for the buttons. 
I aligned with Arjan on the approach here, which of course is open for your feedback. 

What did we do:
- When tabbing trough the interface, the buttons now have a clear 'ring' around them. (see screenshot)
- The `disabled` attribute has been replaced with the `aria-disabled` attribute. This allows all users, both sighted and visually impaired to find the disabled button. Besides, the use of `disabled` on `<a>` is not supported thus wasn't working anyway. More on this [here](https://css-tricks.com/making-disabled-buttons-more-inclusive/) and [here](https://a11y-101.com/development/aria-disabled).

Here's a quick example of the 'ring' shown when tabbing trough the interface:
Schermafbeelding 2021-12-08 om 14 31 49


src/components/ui-button.vue
```diff
@@ -74,7 +79,7 @@ export default {
     },
     created() {
         this.defaultClasses =
-            'leading-6 inline-flex h-11 whitespace-nowrap border-solid rounded-sm align-middle cursor-pointer text-center items-center justify-center focus:outline-none ';
+            'leading-6 inline-flex h-11 whitespace-nowrap border-solid rounded-sm align-middle cursor-pointer text-center items-center justify-center focus:outline-none focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-offset-2 ';
```
bartderijk, line 82: As discussed with Arjan, I'll add an optional offset color to match the parent's background color

src/components/ui-button.vue
```diff
@@ -1,8 +1,13 @@
...
-    <a v-if="href" :href="href" :class="cssClasses" :disabled="disabled">
+    <a v-if="href" :href="href" :class="cssClasses" :aria-disabled="disabled">
         <slot />
     </a>
-    <button v-else :class="cssClasses" :disabled="disabled" v-on="$listeners">
+    <button
+        v-else
+        :class="cssClasses"
+        :aria-disabled="disabled"
+        v-on="$listeners"
```
bartderijk, line 9: If disabled, prevent any events from being triggered as mentioned [here](https://css-tricks.com/making-disabled-buttons-more-inclusive/#:~:text=But%20we%E2%80%99re%20not%20done%20quite%20yet.%20The%20caveat%20here%20is%20that%20we%20still%20need%20to%20prevent%20the%20click%20programmatically%2C%20using%20JavaScript.)

ConnorChristensen:
Thanks for making these changes! 

The only thing that makes me uncomfortable is the disabled link state that uses to `#` as the `href` value.

* This could cause unintended problems if we use section markers within a page (eg `/#listing-3`) and someone clicks a disabled link. That could clear the section marker on the URL.
* If someone is working on filling out a form and they click a disabled link, could it clear the progress they made? Could this clear error states on a form?
* Maybe we can make it a button and set an aria flag that specifies that it is a disabled link?
* Maybe we can set `pointer-events: none` and make the interface a little more usable for a majority of people? This is what the Material UI people do.

I read some of the resources you provided on this pull request and some resources suggested that the button be clickable, but the resulting action be to provide feedback as to why the desired (non-disabled) behavior is not being triggered. If we use the `a` tag with `href="#"` for the disabled state, I think we loose the ability to easily provide the user feedback.

I do not have a "correct" solution here because it feels like we have reached the limit of usability and accessibility in the current web world, and now the decision is how much usability to sacrifice for accessibility and vice versa. If this is overthinking it, please feel free to disregard.

bartderijk:
> Thanks for making these changes!
> 
> The only thing that makes me uncomfortable is the disabled link state that uses to `#` as the `href` value.
> 
> * This could cause unintended problems if we use section markers within a page (eg `/#listing-3`) and someone clicks a disabled link. That could clear the section marker on the URL.
> * If someone is working on filling out a form and they click a disabled link, could it clear the progress they made? Could this clear error states on a form?
> * Maybe we can make it a button and set an aria flag that specifies that it is a disabled link?
> * Maybe we can set `pointer-events: none` and make the interface a little more usable for a majority of people? This is what the Material UI people do.
> 
> I read some of the resources you provided on this pull request and some resources suggested that the button be clickable, but the resulting action be to provide feedback as to why the desired (non-disabled) behavior is not being triggered. If we use the `a` tag with `href="#"` for the disabled state, I think we loose the ability to easily provide the user feedback.
> 
> I do not have a "correct" solution here because it feels like we have reached the limit of usability and accessibility in the current web world, and now the decision is how much usability to sacrifice for accessibility and vice versa. If this is overthinking it, please feel free to disregard.

This is a valid point, I think indeed adding the '#' value can cause some trouble now that you point it out. One suggestion to disabled `a` tags is to remove the `href` attribute. I'll give that a try and see how it works.

arjanvanrees:
Hi @bartderijk, thanks for the great work so far. I'am happy to see that we are able to introduce the ring with a focus state that is only beneficial to keyboard users. And it still looks great. :)

While testing, I discovered some weird thingies. 

Like this one, I notice that this button is styled as a link, but is missing the hover state on the text. So the border color is now the correct color, but the button stays unchanged. If we correct the hover state (this should be blue-0, I think), I think this looks good.
color

The disabled button has a ring, but it doesn't have a color we use. I think this is a Tailwind default. Can we use the grey color of the button for the ring as well? Like we do for all other buttons? What do you think?
disabled

This is a hard one, but those buttons are generally used on top of an image. Using a blue ring would be the best choice here. Could we make the ring white as well?
white

tailwind.config.js
```diff
@@ -9,4 +9,12 @@
 module.exports = {
     presets: [require('@funda/tailwind-config')],
     purge: ['./**/*.vue', './**/*.stories.js'],
+    // @TODO Move this to @funda/tailwind-config once done
```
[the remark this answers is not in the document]
bartderijk: Adding it in this PR https://github.com/funda-frontend/tailwind-config/pull/53

bartderijk:
@ConnorChristensen I went with the sure fire way coined [here on CSS Tricks](https://css-tricks.com/how-to-disable-links/#surefire-way-remove-the-href) which is to not add the `href` attribute when the `<a>` tag is disabled.

bartderijk:
@arjanvanrees Thanks! I'll get cracking with your feedback, will keep you posted.

ConnorChristensen:
> I went with the sure fire way ... which is to not add the `href` attribute when the `<a>` tag is disabled.

I did not know that was an option! Thank you for finding that and adding that in. 

Maybe when I am older I will understand why all these other developers are saying to never make disabled buttons and links, but for now I am happy that our disabled buttons and button links feel well implemented. 😄

bartderijk:
@arjanvanrees I've resolved all issues you mentioned, when you have a moment could you check them out?
@ConnorChristensen I did some refactoring, IMO the computed prop `cssClasses` was doing too many things so I split it up.

arjanvanrees (approved):

dine (approved):

ConnorChristensen (approved):
Approving to bring back approval from Arjan and Dine after the rebase on master.

PR #717, opened 2022-01-06
Add a small badge variant and allow custom colors

bartderijk:
This PR introduces a small variant of the badge component, plus the badge can now come in all the colors of the Funda rainbow 🌈🚀

This UI element is currently seeing use in more and more places, so I figured it was time to add it to the library. I did an inventorisation of the colors used in the designs of the badge and found there's a need for this to be flexible, as there as so many different combinations with new ones being introduced as we speak.


dine (approved):

src/stories/ui-badge.stories.js
```diff
@@ -3,17 +3,39 @@ import UiBadge from '@/components/ui-badge.vue';
...
     title: 'Visual/Badge',
     component: UiBadge,
+    argTypes: {
+        variant: {
+            control: {
+                type: 'select',
+                options: ['default', 'small'],
+            },
+        },
+        backgroundColor: {
```
ConnorChristensen, line 13: Storybook is actually smart enough to add these controls without this (`textColor` also). There is some introspection that goes on behind the scenes, so this is not needed unless you want to add some descriptions or modify the default control type that it generates.

src/stories/ui-badge.stories.js
```diff
@@ -3,17 +3,39 @@ import UiBadge from '@/components/ui-badge.vue';
...
 
-const Template = () => ({
+const Template = (args, { argTypes }) => ({
+    props: Object.keys(argTypes),
     components: { UiBadge },
-    template: `
-        <ui-badge>
-            Badge
-        </ui-badge>`,
+    template: ` <ui-badge :variant="variant" :background-color="backgroundColor" :text-color="textColor">
```
ConnorChristensen, line 29: Remember that you can `v-bind="$props"` to automatically bind all props, which eliminates the risk that a new prop type does not get connected in the story. 😃

src/components/ui-badge.vue
```diff
@@ -1,5 +1,42 @@
...
 </template>
+
+<script>
+export default {
+    props: {
+        variant: {
+            type: String,
+            default: 'default',
+        },
+        backgroundColor: {
```
ConnorChristensen, line 14: I love this interface! Very flexible without just opening up a CSS override class. This is personal preference, but I would include a validator here that ensures that any value passed in matches the regex `bg-[\-a-z\d]+$` (not a perfect regex, but I do not want to be too strict). Of course, anyone will still be able to pass along whatever they want, but Vue will alert them that they are abusing this prop.

src/components/ui-badge.vue
```diff
@@ -1,5 +1,42 @@
...
+            type: String,
+            default: 'bg-blue-6',
+        },
+        textColor: {
```
ConnorChristensen, line 18: See validator comment from `backgroundColor`, as same feedback applies to this (but with regex `^text-.*`)

ConnorChristensen (approved):
I have some suggestions for improvements, but none of them are mandatory, so do with this feedback what you will. 😄

src/stories/ui-badge.stories.js
bartderijk: I tried, but it doesn't pick up on it nicely. Do you have an example?

src/components/ui-badge.vue
bartderijk: I went with a check using `substr` which I think is a little more developer friendly and might be just a bit easier on the performance as well.

ConnorChristensen (approved):
Looks good!

PR #718, opened 2022-01-07
refactor(tooltip): set a better basic example and fix the trigger

bartderijk:
This PR introduces two changes:
- A better basic example of the tooltip, so the arrow is in the middle.
- A valid trigger value, 'click' did not work and was causing an warning.
---
The warning that was resolved:
Schermafbeelding 2022-01-07 om 14 06 02
---
The fixed example (new):
Schermafbeelding 2022-01-07 om 14 02 04
---
The sub-optimal example (previous)
Schermafbeelding 2022-01-07 om 14 01 25
---


ConnorChristensen:
For anyone that stumbles across this in the future, the `click` option for the `trigger` prop is deprecated (in favor of `clickToToggle`), which is why it was throwing an error.

Also, the associated story `trigger` is not reactive to prop changes, so if you want to see the prop change take effect, you need to reload the story with the new `trigger` prop.

ConnorChristensen (approved):
Code changes look good. There is more room for improvement for this story, but we can address that later if you like.

This one may be personal preference, but I would argue that this commit is actually a `chore` rather than a `refactor`. The distinction being that this change has no impact on a person consuming this package, while a refactor likely will have an impact (even if in a minor way).

daniarsuprianto:
I think the triggers can be array, so it should be possible to choose multiple values from the dropdown

bartderijk:
> I think the triggers can be array, so it should be possible to choose multiple values from the dropdown

Indeed, they are. They already were actually.

PR #724, opened 2022-01-10
Add extra small tooltip variant

bartderijk:
This PR introduces the option to display the tooltip using a smaller width.

This fixes the issue where the tooltip is cut off when used in context where there is little space, like in a sidebar as we're currently seeing in funda desk:

Schermafbeelding 2022-01-10 om 11 33 39
Here's an example of the issue that would be resolved by adding `extra-small` to the component, which is introduced with this PR.


src/components/ui-tooltip.vue
```diff
@@ -42,8 +42,21 @@ export default {
             type: Object,
             default: () => {},
         },
+        extraSmall: {
```
daniarsuprianto, line 45: I'd rather have size variant instead of an `extraSmall` prop. I think it's more scalable that way, otherwise if we want to have "large" tooltip, you'd need to add another boolean prop `large` or `extraLarge`

daniarsuprianto (changes requested):

src/components/ui-tooltip.vue
bartderijk: I considered that, but there is currently no requirement for adding large tooltips. I doubt there ever will be, so let's cross that bridge when we find it. By having it like as a boolean, we can add it nice and clean like `extra-small` on the component.

daniarsuprianto: true that now it's not needed and we should just add something that is required for now. but i would still go for more scalable option rather than going through some refactoring and breaking changes

ConnorChristensen: I agree with Daniar. Even if there is no plan for other variants in the future, this is a boolean to control a non-boolean value such as width. 

I have not worked with the tooltip component before, so feel free to correct me here, but intuitively, the tooltip should either:

1. exist as a floating element that is independent of everything on the page except for the reference
2. respect the size of the parent node and constrain the width to fit within the parent

If you have an explanation as to why both of these options are not technically possible, then we can think about what interface this component should take to be able to manually control the size.

ConnorChristensen: I agree with Daniar. Even if there is no plan for other variants in the future, this is a boolean to control a non-boolean value such as width.

I have not worked with the tooltip component before, so feel free to correct me here, but intuitively, the tooltip should either:

1. exist as a floating element that is independent of everything on the page except for the reference
2. respect the size of the parent node and constrain the width to fit within the parent

I think those options should be addressed first. If you have an explanation as to why both of these options are not technically possible, then we can move forward with an interface for manually controlling the maximum width.

bartderijk: Excellent point @ConnorChristensen Indeed it should, and as it turns out it works just like that already.
What was happening:
- The tooltip was used in a sidebar, set to `position: sticky`;
- This tooltip itself is set with `position: absolute`
- This means it was set relative to the sidebar, causing incorrect sizing.
How it was fixed:
- The parent of the tooltip was set to `position: relative`.
- The tooltip now sets the width relatively to that context, sized to measure.

PR #817, opened 2022-04-05
Add listener to anchor tag

daniarsuprianto:
Event listeners were only attached to the `button` tag. Therefore when we're using the anchor tag from the parent and listen to the click event, it didn't work.
Example:
NOW
```
// this doesnt work
<ui-button href="/" @click="$emit('click')"/>
// this works
<ui-button @click="$emit('click')"/>
```
AFTER THIS PR
```
// both work ✌🏼
<ui-button href="/" @click="$emit('click')"/>
<ui-button @click="$emit('click')"/>
```


Zishrodrigues (approved):

src/components/ui-button.vue
```diff
@@ -5,6 +5,7 @@
         :class="cssClasses"
         :aria-disabled="disabled"
         tabindex="0"
+        v-on="inputListeners"
```
ivana-at-funda, line 8: <3

ivana-at-funda (approved):

PR #839, opened 2022-04-26
fix(toaster): avoid duplicated keys

dine:
Fix for this warn error:
Screenshot 2022-04-26 at 08 35 05


ivanasetiawan (approved):
Thank you Dine 🙏

PR #849, opened 2022-04-28
fix(build): remove sideEffects false

dine:


renancarvalho (approved):

jpsc:
It would be nicer to get some sort of context for the change. It helps future questions about the change.

PR #861, opened 2022-05-12
fix(switch): set default checked value to false

ivanasetiawan:
BREAKING CHANGE: Previously we used to expect the default checked value of switch to be true, now it's false


ConnorChristensen:
Can you add an explanation as to why this change is being made?

dine (approved):

dine:
> Can you add an explanation as to why this change is being made?

I think boolean props are usually expected to be false by default. In this case the toggle has an animation, and being by default "on", usually loads and then automatically animates to "off" after fetching the data for example. What do you think?

ivanasetiawan:
@ConnorChristensen: 
We had an issue with the switch component "jumping" from checked to unchecked when a value of a field is `false`. Our team thinks it's a weird behavior. It's OK if the state jumps from false to true, but the other way around feels a bit unintended. 

@dine and I both agree that the default state should be `false`. We cannot assume by default that users say "yes" to whatever fields we ask them.

ConnorChristensen (approved):
Sounds good! I think this is not a breaking change, just a bug fix, but either way is fine. Just like to be sure why a change is being made. 😄

dine:
> Sounds good! I think this is not a breaking change, just a bug fix, but either way is fine. Just like to be sure why a change is being made. 😄

Yeah, but just in case someone was using this bug as a "feature" 😄

PR #927, opened 2022-06-16
feat(breadcrumbs): add breadcrumbs component

dine:


src/components/ui-breadcrumbs.vue
```diff
@@ -0,0 +1,46 @@
+<template>
+    <ul v-if="items.length > 0" class="flex overflow-x-scroll">
+        <li
+            v-for="(item, index) in items"
+            :key="index"
+            :class="{ 'text-dark-3': index === items.length - 1 }"
+        >
+            <span
+                v-if="index === items.length - 1"
```
Zishrodrigues, line 9: This check is done twice in the template. Consider putting it in a `computed` prop.

Zishrodrigues (approved):

src/components/ui-breadcrumbs.vue
jpsc: Also you can use `items.at(-1)`. [.at MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/at)

src/components/ui-breadcrumbs.vue
```diff
@@ -0,0 +1,46 @@
...
+                    whitespace-nowrap
+                    align-middle
+                    cursor-pointer
+                    items-center
+                    focus:outline-none
+                    text-dark-2
+                    hover:text-dark-1
+                "
+                :href="item.href"
+                >{{ item.label
```
jpsc, line 30: Do we need to care about long text by using css truncate or something similar?

src/components/ui-breadcrumbs.vue
```diff
@@ -0,0 +1,46 @@
...
+                }}<IconArrowRight
+                    role="presentation"
```
jpsc, line 32: The svg itself already includes a `role="img"` so I'm not sure if this is valid.

jpsc:
I'm not sure if this component has the correct API. The way that is built doesn't allow for the use of a `<NuxtLink>` component. I would argue that a big part of our breadcrumbs would work inside of the same `Nuxt` application.

PR #930, opened 2022-06-17
Feature/carousel

Zishrodrigues:
Carousel component
- optional controls
- ability to disable & destroy carousel
- separate slide component


jpsc:
@Zishrodrigues what are the exact requirements for our carousel? It seems too big of a dependency (140kb, 38gz)

Can't we, with a more limited set of functionality, create a more basic implementation using `scroll-behavior: smooth` and [scroll-snap-type: x mandatory](https://developer.mozilla.org/en-US/docs/Web/CSS/scroll-snap-type) ?

src/components/ui-carousel.vue
```diff
@@ -0,0 +1,138 @@
+<template>
+    <div class="relative">
+        <ui-button
```
jpsc, line 3: Both these buttons seem to share almost everything except some classes. I would exact that to a carousel controls component.

src/components/ui-carousel.vue
```diff
@@ -0,0 +1,138 @@
...
+                slidesPerView: 3,
+                slidesPerGroup: 3,
+                freeMode: true,
+            }),
+        },
+        hasControls: {
+            type: Boolean,
+            default: true,
+        },
+        enabled: {
```
jpsc, line 91: What is a use case for initializing the carousel component not enabled?

Zishrodrigues: The use case is something like this: https://git.funda.nl/projects/WEBAPPLICATIONS/repos/funda.website/browse/Funda.Static/src/vue/containers/LocalInsightsRelated.vue#112

On mobile the carousel is simply a vertical scroll without any functionality. But in a case where a user increases (or decreases) viewport width on runtime you want to be able to enable/disable the carousel.

Zishrodrigues:
> @Zishrodrigues what are the exact requirements for our carousel? It seems too big of a dependency (140kb, 38gz)
> 
> Can't we, with a more limited set of functionality, create a more basic implementation using `scroll-behavior: smooth` and [scroll-snap-type: x mandatory](https://developer.mozilla.org/en-US/docs/Web/CSS/scroll-snap-type) ?

It needs to have controls and the ability for loop scrolling. Using something like `scroll-snap-type` to create our own carousel will still require quite some custom JS. It also needs to be maintained and accessible. I think we want to avoid building these kind of components from scratch if there are so many maintained tools we can use.

What's your opinion on a more lightweight package? @dine suggested https://github.com/bartdominiak/vue-snap for example.

jpsc:
> > @Zishrodrigues what are the exact requirements for our carousel? It seems too big of a dependency (140kb, 38gz)
> > Can't we, with a more limited set of functionality, create a more basic implementation using `scroll-behavior: smooth` and [scroll-snap-type: x mandatory](https://developer.mozilla.org/en-US/docs/Web/CSS/scroll-snap-type) ?
> 
> It needs to have controls and the ability for loop scrolling. Using something like `scroll-snap-type` to create our own carousel will still require quite some custom JS. It also needs to be maintained and accessible. I think we want to avoid building this kind of component from scratch if there are maintained tools we can use.
> 
> I've looked around for smaller, lightweight packages to use for this. But the ones I'm coming across are not well maintained, `Swiper` seems to be the most reliable. Also when in our bundle and imported the size Is `66kb`. Which imo is not bad.
> 
> Do you have any suggestions? We can also ask ourselves if we actually want the carousel to be in Funda-ui if it's not used by most apps. A package like `Swiper` is simple enough to just use within the app that needs it.

Oh, I totally agree that we should not be building this if the requirements are actually these. I'm talking about balancing the requirements with load size. Loop scrolling sounds like something to not actually be needed (just an example).

I would actually love to see data of user engagement on the carousel vs random items shown on page load with a see more button. It would probably be effective in certain lists but as an aside from the main content, it would probably not make a difference.

PR #931, opened 2022-06-17
feat(breadcrumbs): add breadcrumbs component

dine:


jpsc:
I had to try it 🫣
An [example using custom render](https://codesandbox.io/s/vue-2-playground-forked-p1ckp3?file=%2Fsrc%2FApp.vue), let me know what you think.

Zishrodrigues (dismissed):

Zishrodrigues (approved):

src/__tests__/ui-breadcrumbs.test.js
```diff
@@ -0,0 +1,63 @@
...
+        ol = wrapper.find('ol');
+        li = ol.findAll('li');
+    });
+
+    it('Should show all items', () => {
+        expect(li.length).toBe(5);
+    });
+
+    it('Should show the links', () => {
+        expect(li.at(1).text()).toBe(itemsMock[1].text);
```
jpsc, line 55: This is a good practical example of why I like using `testing-library`, to query this element you could use something like: 

```js
expect(getByRole('link')).toHaveAttribute('href', '/previous-page')
// or
expect(getByText('Previous Page')).toHaveAttribute('href', '/previous-page')
``` 
Or even both. But we are not using `testing-library` on this project 🥸

It just seems, that checking for the HTML tag as string, seems brittle and would break at a minimal change.

jpsc (dismissed):

src/__tests__/ui-breadcrumbs.test.js
dine: Apparently we do have it installed, we just did not use it until now https://github.com/funda-frontend/ui/pull/931/commits/2ab179b714315898574e886d9da7faf81dd385e5

src/__tests__/ui-breadcrumbs.test.js
```diff
@@ -38,26 +37,35 @@ describe('UiBreadcrumbs', () => {
...
+        getByText(itemsMock[lastIndex].text);
     });
 
-    it('Should show the links', () => {
-        expect(li.at(1).text()).toBe(itemsMock[1].text);
-        expect(li.at(1).html()).toContain('<a');
+    test('Should show the links', () => {
+        const { getAllByRole } = wrapper;
+        expect(getAllByRole('link').length).toBe(4);
+        const randomIndex = Math.floor(Math.random() * (itemsMock.length - 1));
```
jpsc, line 57: 🤣

jpsc (approved):

